PR #3, opened 2016-07-13
Add js linter

samuelli:


package.json
```diff
@@ -0,0 +1,12 @@
+{
```
dstockwell: npm docs say 

> Consider also setting "private": true to prevent accidental publication.

dstockwell:
lgtm

PR #29, opened 2016-07-29
Create scroller div inside app-drawer

keanulee:
To address https://github.com/PolymerElements/app-layout/issues/275


samuelli:
LGTM

PR #39, opened 2016-08-04
Use code mirror instead of prism to allow for editable demos

samuelli:


samuelli:
Linter fails to find CodeMirror, not sure how to fix or suppress that.

PR #43, opened 2016-08-05
Lazy load editor

samuelli:
Static code while the editor loads and animate colorize when it loads.


client/src/code-editor.html
```diff
@@ -6,12 +6,18 @@
   <link rel="stylesheet" href="../bower_components/codemirror/lib/codemirror.css">
   <link rel="import" type="css" href="../bower_components/codemirror/lib/codemirror.css">
   <style>
+    :host[unresolved] #container .CodeMirror {
+      -webkit-filter: grayscale(100%);
```
dstockwell: add unprefixed `filter` too

PR #49, opened 2016-08-09
Add a lazy load system which blocks lazy loading

samuelli:


client/src/lazy-block.html
```diff
@@ -0,0 +1,54 @@
+<link rel="import" href="../bower_components/polymer/polymer.html">
+
+<dom-module id="lazy-block">
+  <script>
+    (function() {
+      var blockers = new Map();
```
dstockwell: `Map` probably isn't supported everywhere we care about.

samuelli: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Map#Browser_compatibility

PR #50, opened 2016-08-10
Change load balancing to use pubsub correctly, instead of the weird way I was doing it.

andymutton:


package.json
```diff
@@ -5,6 +5,9 @@
   "repository": {
     "type": "git"
   },
+  "scripts": {
+    "install": "pip install -t lib -r requirements.txt"
```
dstockwell: add something to the readme about this

dstockwell:
lgtm

package.json
samuelli: This is just `npm install` I think

PR #52, opened 2016-08-11
Add stackdriver instance monitoring to auto run script

andymutton:
Apologies for all of the commits... Squash!!


andymutton:
Why does Travis crash in Selenium all the time? What is that dude's problem!!?

PR #53, opened 2016-08-11
Add basic analytics implementation to catalog

andymutton:


samuelli:
Lgtm. Does the registerElement line work in non supporting browsers? We do import the web components polyfill.

samuelli:
Lint errors, might want to just disable it somehow.

PR #54, opened 2016-08-11
Add version fallback to ingestion

shans:
If the newest library version can't be ingested successfully, fallback on the next-most-recent version so that we have something in the search index.

### DIFF
--- a/src/manage.py
+++ b/src/manage.py
@@ -118,16 +118,18 @@ class IngestVersion(webapp2.RequestHandler):
     response = urlfetch.fetch(util.content_url(owner, repo, version, 'README.md'))
     readme = response.content
 
+    def error(error_string):
+      ver = key.get()
+      ver.error = error_string
+      ver.put()
+      self.response.set_status(200)
+
     try:
       content = Content(parent=key, id='readme', content=readme)
       content.etag = response.headers.get('ETag', None)
       content.put()
     except db.BadValueError:
-      ver = key.get()
-      ver.error = "Could not store README.md as a utf-8 string"
-      ver.put()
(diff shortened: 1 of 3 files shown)


src/manage.py
```diff
@@ -82,15 +80,31 @@ def get(self, owner, repo, kind):
...
+        library.tags_etag = response.headers.get('ETag', None)
+        data.reverse()
+        is_newest = True
         for version in data:
           tag = version['ref'][10:]
           if not versiontag.is_valid(tag):
             continue
           sha = version['object']['sha']
+          params = {}
+          if (is_newest):
```
dstockwell: excess parens

src/manage.py
```diff
@@ -82,15 +80,31 @@ def get(self, owner, repo, kind):
           github.release()
           library.put()
           return
+        data = [d for d in data if versiontag.is_valid(d['ref'][10:])]
+        if len(data) is 0:
```
dstockwell: merge these two error cases?, set data to [] above

src/manage.py
```diff
@@ -137,18 +163,13 @@ def get(self, owner, repo, version):
     try:
       json.loads(response.content)
```
dstockwell: FIXME? We need to abort (return 500) if the markdown request failed.

dstockwell:
lgtm

src/manage.py
shans: No.

src/manage.py
shans: removed

src/manage.py
shans: done

PR #60, opened 2016-08-15
Use log messages for metrics

andymutton:


samuelli:
lgtm. Do we do any sort of linting/tests on analysis?

To use the ESLint that we run, just need to add it to to target src array in the `Gruntfile.js`.

andymutton:
(I enabled lint and fixed all the lint errors)

analysis/ana_log.js
```diff
@@ -0,0 +1,27 @@
...
+      console.log(path);
+    }
+  }
+  static log() {
+    console.log(...arguments);
+  }
+}
+
+module.exports = {
+  Ana: Ana
```
dstockwell: `module.exports = {Ana};`

dstockwell: Although `module.exports = Ana;` is probably all you need, unless you expect to export something else later?

analysis/ana_log.js
```diff
@@ -0,0 +1,27 @@
...
+  static fail(counterPath) {
+    var path = counterPath + "/fail";
+    if (arguments.length > 1) {
+      console.log(path, ...Array.prototype.slice.call(arguments, 1));
+    } else {
+      console.log(path);
+    }
+  }
+  static log() {
+    console.log(...arguments);
```
dstockwell: Do we need to escape this if the first arg ends with `/success` or `/fail`?

PR #63, opened 2016-08-16
Continuous deployment

samuelli:


dstockwell:
Nice! Is there any way to tell which version has been deployed?

lgtm

PR #81, opened 2016-08-22
Add on-demand API which resolves URLs to SHA

samuelli:


src/api.py
```diff
@@ -203,6 +205,39 @@ def post(self):
...
+      # TODO: trigger on demand task
+      return
+
+    # Resolve SHA using these patterns and Github API
+    tail = re.sub(r'/pull/(.*)', r'pull/\1/head', tail)
+    tail = re.sub(r'/tree/(.*)', r'heads/\1', tail)
+    tail = re.sub(r'^$', r'heads/master', tail)
+
+    if not tail:
+      self.response.set_status(404)
```
dstockwell: `400` makes more sense than `404`

samuelli: Done

PR #85, opened 2016-08-23
Basic on demand form

samuelli:
This change adds a front-end for on-demand requests which uses recaptcha to verify a human is requesting to process a Github URL.

Also includes a bunch of fixes:
 * ESLint all files again
 * Always include versions in URLs
 * Update README.md
 * Specify a target for on-demand task
 * Allow SHAs to be used instead of versions on the element page

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,12 @@ To increase Github API quota, acqure a Github token and store it:
 ```bash
 cat > secrets.yaml
 github_token: 'your-github-token'
-^D
+```
+
+If you would like to use reCAPTCHA, obtain a token and store it:
+```bash
+cat > secrets.yaml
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -21,7 +21,12 @@ To increase Github API quota, acqure a Github token and store it:
 ```bash
 cat > secrets.yaml
 github_token: 'your-github-token'
-^D
+```
+
+If you would like to use reCAPTCHA, obtain a token and store it:
```
dstockwell: Realised it won't work, as there is a matching key in `client/src/catalog-ondemand.html`

dstockwell:
lgtm

PR #95, opened 2016-08-29
Add thread safe and hope that it's true.

andymutton:


andymutton:
Also changed the default start script to point to staging (when deployed it only uses the environment variable and ignores that project arg.)

PR #97, opened 2016-08-30
Add error description when dependencies cant be cound

dstockwell:


src/user_content.py
```diff
@@ -22,6 +22,7 @@ def get(self, owner, repo, tag, name=None, path=None):
```
samuelli: It doesn't fail here?

dstockwell: > It doesn't fail here?

Done!

samuelli:
lgtm, thanks!

PR #98, opened 2016-08-30
Use same validation logic for all mutation requests

dstockwell:


samuelli:
Does this affect API? Or those are fine because its posted as a task?

dstockwell:
The tasks posted by API should still work.

samuelli:
Ok thanks. LGTM.

PR #105, opened 2016-09-06
Make staging always serve demos from user-content-dot-custom-elements-staging.appspot.com

andymutton:


client/src/catalog-app.html
```diff
@@ -70,9 +70,13 @@
         }
       },
 
+      isStaging: function() {
```
dstockwell: Maybe we should just invert all this and use `!isProduction()`. Don't really need to distinguish between staging/localhost then.

dstockwell:
lgtm

PR #106, opened 2016-09-06
Ingest repo stats/participation

dstockwell:


src/manage.py
```diff
@@ -122,7 +122,20 @@ def update_metadata(self):
       self.library.contributor_count = len(json.loads(response.content))
       self.library_dirty = True
     elif response.status_code != 304:
-      return self.abort('cout not update contributors (%d)' % response.status_code)
```
shans: I'm disappointed that we're losing a sneaky reference to this venerable and much-loved C++ stream

PR #107, opened 2016-09-06
Remove unused search fields

dstockwell:


shans:
Need to also remove datetime import now.

There's some other datetime stuff that you might be able to nuke too.

PR #109, opened 2016-09-06
Rework collection references as separate entities

dstockwell:
- [x] move `CollectionReference` from `Library` to standalone entity
- [x] filter collection list in metadata to only one entry per collection
- [x] write tests for:
 - [x] collection ingestion
 - [x] deleting stale collection references

### DIFF
--- a/src/datamodel.py
+++ b/src/datamodel.py
@@ -124,8 +124,6 @@ class Version(ndb.Model):
         existing_version = result_map.get(collection_id, None)
         if existing_version is None or versiontag.compare(collection_version.key.id(), existing_version.key.id()) > 0:
           result_map[collection_id] = collection_version
-      else:
-        raise Exception([version_key.id(), collection_references[i].semver])
     raise ndb.Return(result_map.values())
 
 class Content(ndb.Model):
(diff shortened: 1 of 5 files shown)


dstockwell:
Still TODO:
- [ ] filter collection list in metadata to only one entry per collection
- [ ] write tests for:
 - [ ] collection ingestion
 - [ ] deleting stale collection references

src/datamodel.py
```diff
@@ -99,12 +102,31 @@ class Version(ndb.Model):
...
+      if collection_version is None:
+        # Remove the stale reference.
+        yield collection_references[i].key.delete_async()
+      elif versiontag.match(version_key.id(), collection_references[i].semver):
+        collection_id = collection_version.key.parent().id()
+        existing_version = result_map.get(collection_id, None)
+        if existing_version is None or versiontag.compare(collection_version.key.id(), existing_version.key.id()) > 0:
+          result_map[collection_id] = collection_version
+      else:
+        raise Exception([version_key.id(), collection_references[i].semver])
```
shans: ?

shans:
lgtm

src/datamodel.py
dstockwell: > ?

Oops. Removed and added test.

PR #110, opened 2016-09-06
Centralize github logic and handle github errors gracefully

dstockwell:
This ensures that we're using the same pattern throughout manage -- github server errors bubble up to the top of the request handler where they're converted into a `502` causing the task to be retried.


shans:
LGTM after adding some docs to RequestHandler

PR #117, opened 2016-09-08
Compare new/old tags and only ingest new/updated versions

dstockwell:


src/manage.py
```diff
@@ -197,20 +197,30 @@ def update_metadata(self):
...
-    if latest:
-      params["latestVersion"] = "True"
     version_object = Version.get_or_insert(tag, parent=self.library.key, sha=sha)
-    if url is not None:
-      version_object.url = url
+    if version_object.sha == sha and version_object.status == Status.ready:
+      # Version object is already up to date
+      return
+
+    version_object.url = url
```
andymutton: Do we not care whether the url is None anymore?

dstockwell: I just realized the check was mostly useless. Setting `url = None` is fine.

andymutton:
lgtm

PR #130, opened 2016-09-14
Rework library ingestion and indexing

dstockwell:
Library ingestion is now transactional
* version ingestion tasks won't start until the library has been committed. This avoids attempting to index a version before the library has been written.

Indexing and collection references are now built in a separate task
* removes the need to thread `latest_version` through each version ingestion
* ensures that we actually index the latest version when there are multiple ingestions happening concurrently

Some other cleanups:
* added a utility to build a Library id, `Library.id`
* added a utility to find the latest version for a Library, `Library.latest_version_for_key_async`

### DIFF
--- a/src/datamodel.py
+++ b/src/datamodel.py
@@ -61,6 +61,10 @@ class Library(ndb.Model):
   updated = ndb.DateTimeProperty(auto_now=True)
 
   @staticmethod
+  def id(owner, repo):
+    return '%s/%s' % (owner.lower(), repo.lower())
+
+  @staticmethod
   def maybe_create_with_kind(owner, repo, kind):
     library = Library.get_or_insert('%s/%s' % (owner, repo))
     # FIXME: Probably don't want libraries to change kind.
@@ -77,6 +81,14 @@ class Library(ndb.Model):
     if version_cache is not None:
       versions = version_cache.versions
     raise ndb.Return(versions)
+
+  @staticmethod
+  @ndb.tasklet
+  def latest_version_for_key_async(key):
+    versions = yield Library.versions_for_key_async(key)
+    if versions == []:
+      raise ndb.Return(None)
+    raise ndb.Return(versions[-1])
 
   @staticmethod
   def uncached_versions_for_key(key):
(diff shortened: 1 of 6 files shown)


src/datamodel.py
```diff
@@ -80,28 +84,37 @@ def versions_for_key_async(key):
...
-    if create:
-      version_cache = yield VersionCache.get_or_insert_async('versions', parent=library_key, versions=versions)
-    else:
-      version_cache = yield VersionCache.get_by_id_async('versions', parent=library_key)
-    if version_cache is not None and version_cache.versions != versions:
+  def update(library_key):
+    """Returns whether the latest version has changed and an index update is needed.
+    """
+    versions = Library.uncached_versions_for_key(library_key)
+    version_cache = VersionCache.get_or_insert('versions', parent=library_key, versions=[])
```
samuelli: Remove versions param

src/datamodel.py
```diff
@@ -80,28 +84,37 @@ def versions_for_key_async(key):
...
   @ndb.transactional
-  def update_async(library_key, create=False):
-    versions = yield Library.uncached_versions_for_key_async(library_key)
-    if create:
-      version_cache = yield VersionCache.get_or_insert_async('versions', parent=library_key, versions=versions)
-    else:
-      version_cache = yield VersionCache.get_by_id_async('versions', parent=library_key)
-    if version_cache is not None and version_cache.versions != versions:
+  def update(library_key):
+    """Returns whether the latest version has changed and an index update is needed.
```
samuelli: Updates version and returns...

src/manage.py
```diff
@@ -79,24 +79,47 @@ class RequestHandler(webapp2.RequestHandler):
   Stashing a permanent error in a datastore entity.
   """
```
samuelli: Can we rename the `abort` function to `retry`?

samuelli:
lgtm

src/manage.py
dstockwell: will do in another change

PR #134, opened 2016-09-14
Modify collection ingest/update to track the master branch

dstockwell:
Collections are now detected by the presence of
`"keywords": ["element-collection"]` in `master/bower.json`.

Library entities for collections maintain a sequence number
that is incremented each time we detect that the master branch
has changed. The sequence number is used to construct
artificial version tags which can be compared appropriately.

### DIFF
--- a/src/manage_test.py
+++ b/src/manage_test.py
@@ -232,8 +232,8 @@ class IngestLibraryTest(ManageTestBase):
     library_key = Library(id='org/repo', metadata='{"full_name": "NSS Bob", "stargazers_count": 420, "subscribers_count": 419, "forks": 418, "updated_at": "2011-8-10T13:47:12Z"}').put()
     Version(id='v1.0.0', parent=library_key, sha='sha').put()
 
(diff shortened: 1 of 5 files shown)


src/manage_test.py
```diff
@@ -108,7 +108,35 @@ def test_update_triggers_version_ingestion(self):
...
+    response = self.app.get(util.update_library_task('org/repo'), headers={'X-AppEngine-QueueName': 'default'})
+    self.assertEqual(response.status_int, 200)
+    library = library_key.get()
+    self.assertEqual(library.error, None)
+    self.assertEqual(library.status, Status.ready)
+
+    tasks = self.tasks.get_filtered_tasks()
+    self.assertEqual([
+        util.delete_task('org', 'repo', 'v0.0.1'),
+        util.ingest_version_task('org', 'repo', 'v0.0.2'),
```
samuelli: Is delete conditional on the new version being `Status.ready`?

dstockwell: Not yet. But there are still FIXMEs around doing that.

PR #135, opened 2016-09-14
UI for info bar

samuelli:
image

### DIFF
--- a/client/src/catalog-element.html
+++ b/client/src/catalog-element.html
@@ -12,6 +12,10 @@
 <dom-module id="catalog-element">
   <template>
     <style>
+      * {
+        box-sizing: border-box;
+      }
+
       a {
         text-decoration: none;
       }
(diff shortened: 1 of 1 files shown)


client/src/catalog-element.html
```diff
@@ -12,6 +12,10 @@
 <dom-module id="catalog-element">
   <template>
     <style>
+      * {
```
dstockwell: Does this need to match everything?

client/src/catalog-element.html
```diff
@@ -166,15 +214,34 @@ <h1 id="repo-title">[[repo]]</h1>
...
+          <h3>Dependencies</h3>
+          <template is="dom-repeat" items="[[_dependencies()]]">
+            <div>[[item]]</div>
+          </template>
+        </div>
+        <div id="info-graph">
+          <h3>Latest commit [[_timeSinceString(data.updated_at)]] ago</h3>
+        </div>
+        <div id="info-contributors">
+          <h3>[[data.contributors.length]] contributors</h3>
```
dstockwell: But what if there's only one :P

client/src/catalog-element.html
samuelli: No, but its a pretty common pattern.

client/src/catalog-element.html
samuelli: TODO

samuelli: I'll address this when it moves to an element

PR #149, opened 2016-09-16
Format dependencies in catalog-element

dstockwell:


client/src/catalog-element.html
```diff
@@ -113,6 +112,18 @@
         flex: 0 0 270px !important;
       }
 
+      #info-dependencies {
+        line-height: 32px;
+        color: black;
```
samuelli: ?

client/src/catalog-element.html
```diff
@@ -113,6 +112,18 @@
...
+      }
```
samuelli: \n

PR #154, opened 2016-09-16
Show demos in an overlay dialog

dstockwell:


client/src/catalog-element.html
```diff
@@ -403,6 +423,22 @@ <h2 title>[[subElement]]</h2>
         this._routeChanged(this.route, this.visible);
       },
 
+      _showDemo: function(e) {
```
samuelli: We shouldn't use this approach, its pretty confusing and demos aren't linkable anymore.

Instead let's use the routing that's already set up. In `_routeChanged()` we just need to check `this.demo` and open/close the dialog based on that.

Closing the dialog should also result in a route change.

dstockwell: Done!

client/src/catalog-element.html
```diff
@@ -486,6 +506,37 @@ <h2 title>[[subElement]]</h2>
...
+        }
+      },
+
+      _demoClicked: function(e) {
+        this.pathBeforeDemo = this.route.path;
+      },
+
+      _showDemo: function(e) {
+        var demoUrl = this.baseUrls.userContent + '/' + this.routeMatch.owner + '/' + this.repo + this.versionRoute + '/' + this.repo + '/' + this.demo;
+        this.$['demo-frame'].src = demoUrl
```
samuelli: ;

dstockwell: Done!

PR #185, opened 2016-09-20
Return latest_version in brief metadata, use it to link when necessary

dstockwell:


client/src/catalog-item.html
```diff
@@ -106,6 +106,14 @@
         data: Object
       },
 
+      versionPathIfNecessary: function(version, latestVersion) {
+        if (version != latestVersion) {
+          return '/' + version;
+        } else {
+          return '';
+        }
```
samuelli: ```return latestVersion == version ? '' : '/' + version;```  ?

client/src/catalog-item.html
```diff
@@ -106,6 +106,14 @@
         data: Object
       },
 
+      versionPathIfNecessary: function(version, latestVersion) {
```
samuelli: _versionPath()

PR #189, opened 2016-09-21
Secure task submission to analysis by using a task queue instead of pubsub.

andymutton:
PubSub doesn't support authenticated URLs in Managed VMs (I KNOW, RIGHT?!!) and doesn't actually provide any headers that identify the request as having come from PubSub. CRAZY!
Changing to task queues allows us to check the existence of the x-appengine-queuename header, which means that the request came from our task queue.

- changed main.js to expect task-queue get with params, rather than pubsub json body (and transform it back to attributes for the return journey)
- validate that request comes from task queue and has required params
- set staleness on lockfile so that fatal errors don't leave us with a permanent lock
- submit requests to task queue instead of pubsub
- don't retry requests with an invalid tag/branch
- remove pubsub request subscriptions and topics from config, and delete testing message, update docs


analysis/src/main.js
```diff
@@ -38,17 +37,35 @@ function processTasks() {
...
+    if (req.query.sha) {
+      attributes.sha = req.query.sha;
+    }
+
+    // We only accept requests from the task queue service.
+    // This check is valid because appengine strips external x-appengine headers.
+    // By definition, this is internal.
+    if (!req.get('x-appengine-queuename')) {
+      Ana.fail("main/processTasks/originator-not-appengine", JSON.stringify(attributes));
+      res.status(200).send(); // Don't retry
```
dstockwell: I think `403` makes more sense here. If this ever were to happen we'd probably want to retry anyway.

analysis/README.md
```diff
@@ -17,17 +17,17 @@ rebuild-pubsub.sh [custom-elements, custom-elements-staging]
 
 ## Running it locally
 Make sure you have node and npm installed and then run the following. It will
-use the custom-elements-staging pubsub (but won't receive push requests for
+use the custom-elements-staging pubsub to responde (but won't receive push requests for
```
dstockwell: Oh no! A typo!

src/util.py
```diff
@@ -87,6 +61,11 @@ def ingest_commit_task(owner, repo):
 def ingest_webhook_task(owner, repo):
   return '/task/ingest-webhook/%s/%s' % (owner, repo)
 
+def ingest_analysis_task(owner, repo, version, sha=None):
+  if sha:
```
dstockwell: `if sha is not None`

PR #207, opened 2016-09-22
Send error data back to catalog for permanent failures and update status accordingly.

andymutton:


analysis/src/main.js
```diff
@@ -84,6 +85,8 @@ function processTasks() {
         } else {
           Ana.fail("main/processTasks");
           res.status(200).send();
```
dstockwell: does this still happen if line 89 fails?

src/manage.py
```diff
@@ -650,14 +650,14 @@ def handle_post(self):
     owner = attributes['owner']
     repo = attributes['repo']
     version = attributes['version']
+    error = attributes['error'] if 'error' in attributes else None
```
dstockwell: attributes.get('error', None)

analysis/src/main.js
andymutton: Yup. That's why I ordered it this way.

dstockwell: But shouldn't we retry until we're able to send the result back?

PR #278, opened 2016-09-27
Move navigation progress and fade to catalog-app

dstockwell:


client/src/page-loading-behavior.html
```diff
@@ -0,0 +1,19 @@
+<script>
+window.PageLoadingBehavior = {
+  properties: {
+    loading: {
+      type: Boolean,
+      value: true,
+      observer: '_loadingChanged',
+      notify: true,
```
samuelli: Does it need to be notify true with an observer?

dstockwell: No!

PR #283, opened 2016-09-28
Retry analysis requests if we fail to send failure messages after failing.

andymutton:
also...
- change status(x).send() to sendStatus(x), cause that's a thing
- move property filtering to hydrolysis where it belongs (and keep the more comprehensive hierarchy filtering)


andymutton:
It wasn't my plan, but I'm fine with it. Thanks for noticing.

PR #299, opened 2016-09-29
Allow demos to be opened fullscreen in new tab

### LINKED ISSUE #294: cmd+click on a demo should open up in a new tab without the modal

samuelli:
Fixes #294

### DIFF
--- a/client/src/catalog-element.html
+++ b/client/src/catalog-element.html
@@ -525,7 +525,7 @@
               <a href="[[route.prefix]]/[[owner]]/[[repo]][[versionRoute]]/[[element.is]]">[[_docLabel(element.is)]]</a>
 
               <template is="dom-repeat" items=[[element.demos]] as="demo">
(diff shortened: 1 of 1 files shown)


client/src/catalog-element.html
```diff
@@ -707,7 +707,15 @@ <h2 title>[[subElement]]</h2>
         }
       },
 
-      _demoClicked: function() {
+      _demoClicked: function(event) {
+        var isMac = navigator.userAgent.indexOf('Mac OS X') != -1;
+        var mouseEvent = event.detail.sourceEvent;
+        if ((!isMac && mouseEvent.ctrlKey) || (isMac && mouseEvent.metaKey)) {
```
dstockwell: I think if ctrl/meta/anything is pressed we should just fall back to the default behavior. Seems less fragile than relying on the useragent.

dstockwell:
> dstockwell approved these changes just now

oh, so that's what that button does.

client/src/catalog-element.html
samuelli: Shouldn't we better match browser behavior?

samuelli:
Pending comments on #294

PR #312, opened 2016-09-29
Scroll to top on page changes

### LINKED ISSUE #287: Navigating between collections or between elements does not scroll to top

samuelli:
Fixes #287

### DIFF
--- a/client/src/catalog-app.html
+++ b/client/src/catalog-app.html
@@ -194,6 +194,8 @@
       },
 
       _pageReady: function() {
+        Polymer.AppLayout.scroll(0);
+
         if (this.$.pages.selectedItem && this.$.pages.selectedItem.loading !== true)
           this._clearLoading();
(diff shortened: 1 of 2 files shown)


client/src/catalog-app.html
```diff
@@ -194,6 +194,8 @@
       },
 
       _pageReady: function() {
+        Polymer.AppLayout.scroll(0);
```
dstockwell: I think we should do this in `_setLoading` since `_pageReady` only gets called when `page` changes.

client/src/catalog-collection.html
```diff
@@ -157,6 +158,8 @@ <h1 class="section-header">[[_collections.length]] Collections</h1>
         if (!visible || !this._loaded || route.prefix != '/collection')
           return;
 
+        Polymer.AppLayout.scroll(0);
```
dstockwell: Shouldn't need this with the above.

PR #315, opened 2016-09-29
Show published element after submission

### LINKED ISSUE #289: Publish should show acknowledgement after pressing the button

samuelli:
Fixes #289


client/src/catalog-publish.html
```diff
@@ -60,7 +60,7 @@ <h1 class="page-title">Publish your element</h1>
 
         <input type="text" name="repo" value="{{repo::input}}" placeholder="user/repo">
         <div class="g-recaptcha" data-sitekey="6LcCRSgTAAAAAGhe5bUDS5lVHAPNN2cY7xo-Zk5f"></div>
-        <button on-tap="onSubmit">Publish</button>
+        <a href="/preview/[[repo]]"><button on-click="onSubmit">Publish</button></a>
```
dstockwell: Using `<a>` is a bit weird here. You can right click to open in new window, tab order gets weird, etc. Can we just do this in `onSubmit`?

PR #334, opened 2016-10-04
Favicons

samuelli:
Fixes #334


client/manifest.json
```diff
@@ -1,8 +1,18 @@
...
-  "name": "catalog",
-  "short_name": "catalog",
-  "start_url": "/",
-  "background_color": "#6dc38f",
-  "display": "standalone",
-  "theme_color": "#6dc38f"
+    "name": "Web Components",
+    "icons": [
+        {
+            "src": "\/assets\/android-chrome-192x192.png",
```
dstockwell: why are these called android-chrome?

dstockwell:
Should `client/assets/apple-touch-icon.png` have a transparent background? (it's white currently)

client/manifest.json
samuelli: Only used by android chrome I guess. Default names the tool spit out. Not sure anyone else uses this manifest

dstockwell: We should rename them :P

samuelli:
@dstockwell: No, its intentional. iOS does not support transparent icons. If there is transparency, it is replaced with black.

abdonrd:
I recommend check the [index.html](https://github.com/PolymerElements/polymer-starter-kit/blob/master/index.html) and [manifest.json](https://github.com/PolymerElements/polymer-starter-kit/blob/master/manifest.json) from Polymer Starter Kit.

Here a discussion about that: [Update to consistent index.html and manifest.json](https://github.com/Polymer/app-drawer-template/pull/13)

PR #339, opened 2016-10-04
Support multiple inline demos

### LINKED ISSUE #327: Support multiple inline demos

samuelli:
Fixes #327


dstockwell:
lgtm, but message channel would let us avoid the global onmessage handlers

PR #349, opened 2016-10-05
Move search out of the dialog

dstockwell:


client/polymer.json
```diff
@@ -11,7 +11,6 @@
     "src/catalog-publish.html",
     "src/catalog-home.html",
     "src/lazy-resources.html",
```
samuelli: Invalid json

PR #364, opened 2016-10-05
Add README to publishing requirements

katejeffreys:


samuelli:
I'm not sure we need to include this. It's standard practice on GitHub.

katejeffreys:
Ok. I don't mind either way. @tjsavage wanted it to be included - please see his comment in the task list I created.

https://docs.google.com/a/google.com/spreadsheets/d/1H8TWNWwPqXzpmmoVYg7a8f2TjEyfuO8nGYiCo4KZtD0/edit?usp=sharing

tjsavage:
I don't think it would hurt to mention that the primary thing that's going to be shown on the catalog is the README.md file. Figure though it's common GH practice, it might not be immediately obvious that that's where the content is coming from, and might save some devs a round-trip.

PR #413, opened 2016-10-10
Replace iron doc viewer with something that looks a lot like it.

andymutton:


client/src/catalog-api-property.html
```diff
@@ -0,0 +1,311 @@
...
+
+        /**
+         * Resets any state that was set up for transitions.
+         *
+         * We are careful to reset our explicit heights after a transition
+         * completes, so that the property doesn't clip values if the user resizes
+         * their viewport.
+         * @param {object} event The event in question.
+         */
+        _onTransitionEnd: function(event) {
```
dstockwell: A bunch of these functions don't seem to be used.

client/src/catalog-api-property.html
```diff
@@ -0,0 +1,311 @@
...
+        display: none;
+      }
+
+      .deeplink {
+        color: currentcolor;
+        text-decoration: none;
+      }
+
+      .deeplink:hover {
+        color: var(--paper-pink-500);
```
dstockwell: We should try to avoid paper colors.

client/src/catalog-api-property.html
andymutton: I've deleted a bunch of functions.

client/src/catalog-api-property.html
andymutton: Doned it.

PR #434, opened 2016-10-11
Add suppressed status

### LINKED ISSUE #250: API/datamodel/manage: Add suppress field to block api results and ingest/update for a specific library.

dstockwell:
Fixes #250


dstockwell:
> How will something get suppressed? Not convinced we should suppress all deletes

Manually in the console. This should only be used to prevent re-ingestion. Can use `manage/delete` in many cases.

PR #456, opened 2016-10-11
Flesh out the section on documenting your element

katejeffreys:


samuelli:
Need to use spaces instead of tabs

Issue #416, opened 2016-10-10
Searching for "check" has no results but "checkbox" does


santokii:
Because a lot of titles have phrases with no sentence breaks, maybe add the ability to search for words within phrases or titles that may not be separated out? Also, on the reverse side some words like datepicker may be searched as 1 word as apposed as 2.

andymutton:
'check' now returns what it should (as does 'che', 'chec', 'checkb' ... ). 'box' does not, but it's not clear that introducing word-splitting/dictionary matching is a sensible option here.

Title and camel case splits are also taken care of (eg. 'che', chec', 'box', 'but', 'butt', 'butto' will match CheckBoxButton)

Prefix based indexing and matching should solve most search issues, so closing this pending additional need.

PR #460, opened 2016-10-12
Improve search index

andymutton:
Addresses some of #416, by adding prefix matching on descriptions and repo.
Arbitrary substring indexing would probably blow the index (and lower quality), but dictionary-based substrings might be worthwhile.


src/manage.py
```diff
@@ -660,16 +660,20 @@ def update_collection_dependencies(self, collection_version_key, bower):
       task_url = util.ensure_library_task(dep.owner.lower(), dep.repo.lower())
       util.new_task(task_url, target='manage')
 
+
   def update_search_index(self, owner, repo, version_key, library, bower):
     metadata = json.loads(library.metadata)
     fields = [
-        search.TextField(name='owner', value=owner),
+        search.AtomField(name='owner', value=owner),
```
dstockwell: Fixes https://github.com/webcomponents/beta/issues/281

PR #469, opened 2016-10-12
Fill out user author profiles

samuelli:


samuelli:
"Collected by" is too big in search focus

PR #471, opened 2016-10-12
Fix some loading issues with catalog-list

dstockwell:


client/src/catalog-list.html
```diff
@@ -80,7 +83,7 @@ <h1 class="section-header">[[_collections.count]] Collections</h1>
       </div>
     </template>
     <template is="dom-if" if="[[_elements.count]]"><h1 class="section-header">[[_elements.count]] Elements</h1></template>
-    <div id="elements" class="box loader" hidden="[[!_elements.count]]" loading$="[[!_elements]]">
+    <div id="elements" class="box loader" hidden="[[!_elements.count]]" loading$="[[!_and(_elements, _collections)]]">
```
samuelli: Is this just to make sure it appears second?

dstockwell: Yes.

PR #477, opened 2016-10-12
Filter routes before passing to author/collection/search

dstockwell:


client/src/catalog-collection.html
```diff
@@ -123,48 +116,29 @@ <h2 class="page-subtitle">[[data.description]]</h2>
...
-        if (disabled)
+      _pathChanged: function(path, visible) {
+        if (!visible || !path || path == this._currentPath)
           return;
 
-        document.title = this._routeMatch.repo;
+        document.title = this._repo;
+        this._currentPath = path;
 
+        this._fullName = path.substring(1);
```
samuelli: Change this to strip the / as well, and use `path` instead of `subroute`

client/src/catalog-list.html
```diff
@@ -174,6 +181,7 @@ <h1 class="section-header">[[_collections.count]] Collections</h1>
           this.notifyPath('_elements.results');
         }
         this.$.threshold.clearTriggers();
+        this.$.threshold.checkScrollThesholds()
```
samuelli: ;

client/src/catalog-search.html
```diff
@@ -37,21 +37,12 @@ <h1 class="page-title">[[_title(_query, title)]]</h1>
...
 
-      pathChanged: function(path, visible) {
-        var disabled = !visible || this._currentPath == path;
-        this.$.list.disabled = disabled;
-        if (disabled)
-          return;
-        this._currentPath = path;
-        this._query = path.substring(1);
+      _query: function(path) {
+        return path && path.substring(1);
```
samuelli: Strip it here too? This is already path

client/src/catalog-collection.html
dstockwell: Done

PR #503, opened 2016-10-13
Hide repo contents when there are no options

samuelli:


samuelli:
This doesn't animate properly because the items are now display: hidden; which means they go straight to `opacity: 1` bypassing `opacity: 0`

PR #507, opened 2016-10-13
Clean up demo navigation and catalog-element loading

### LINKED ISSUE #245: Client: Dismissing demos should revert history to before the demo was opened

dstockwell:
* Back/forward button navigates in and out of demos
* Dismissing a demo always reveals the underlying element's page
* Version is no longer required in demo path

Fixes #245

### DIFF
--- a/client/src/catalog-element.html
+++ b/client/src/catalog-element.html
@@ -555,7 +555,7 @@
             <a href="[[route.prefix]]/[[owner]]/[[repo]][[versionRoute]]">Overview</a>
 
             <template is="dom-repeat" items="[[dupedDemos]]" as="demo">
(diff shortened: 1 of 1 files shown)


client/src/catalog-element.html
```diff
@@ -726,54 +725,75 @@ <h1 hidden="[[subElement]]">README.md</h1>
...
+        if (!visible || !(prefix == '/element' || prefix == '/preview'))
           return;
 
         if (prefix == '/preview' && path == '')
           return;
 
+        if (path == this._currentPath)
+          return;
+
+        this._currentPath = path;
```
samuelli: this._cached = this.$.metaAjax.url

PR #509, opened 2016-10-13
Make SW only cache 1XX/2XX/3XX responses

### LINKED ISSUE #453: Preview: element continues to fail preview process
[Origin page](https://beta.webcomponents.org/preview)

There's no way to get back to the previous page once an element  fails to preview. There should be a "Re-try" button or some.

screen shot 2016-10-11 at 11 48 07 am

dstockwell:
Fixes #453


dstockwell:
I'll test this on staging before merging.

PR #514, opened 2016-10-14
Load more collections on scroll (if there's no elements on the page).

andymutton:


client/src/catalog-list.html
```diff
@@ -68,15 +68,27 @@
...
+        <!-- Don't autoload more collections if there are elements -->
+        <template is="dom-if" if="[[_elements.count]]">
+          <template is="dom-repeat" items="[[_collections.results]]">
+            <collection-card class="box" data="[[item]]"></collection-card>
+          </template>
+          <div id="spacer"></div>
+        </template>
+        <!-- But if there aren't, go for it, it's the 90s -->
+        <template is="dom-if" if="[[!_elements.count]]">
+          <iron-scroll-threshold on-lower-threshold="_loadMoreCollections" id="threshold" scroll-target="document" lower-threshold="800">
```
dstockwell: hmm, I think we should move the logic to a _loadMore function instead, it would decide which one to extend. we probably don't need to change the template at all then.

andymutton: Done. PTAL.

PR #517, opened 2016-10-14
Add syntax highlighter everywhere

samuelli:


client/src/catalog-static.html
```diff
@@ -114,12 +116,28 @@ <h1 class="page-title">Introduction</h1>
...
       },
 
+      _visibilityChanged: function(visible) {
+        if (!visible)
+          return;
+        Polymer.RenderStatus.afterNextRender(this, function() {
+          var codeBlocks = Polymer.dom(this.root).querySelectorAll('pre[lang]');
+          for (var i = 0; i < codeBlocks.length; i++)
+            this.scopeSubtree(codeBlocks[i], true);
+          this.fire('syntax-highlight', {codeBlocks: codeBlocks});
```
dstockwell: Is it OK to do this more than once?

client/src/syntax-highlighter.html
```diff
@@ -0,0 +1,85 @@
...
+      };
+
+      if (map[lang])
+        lang = map[lang];
+
+      return lang || 'htmlmixed';
+    },
+
+    _highlightElement: function(element) {
+      if (element.__syntaxHighlighted)
```
dstockwell: Seems OK then.

client/src/catalog-syntax-styles.html
```diff
@@ -0,0 +1,79 @@
+<link rel="import" href="../bower_components/polymer/polymer.html">
+
+<dom-module id="catalog-syntax-styles">
+  <link rel="stylesheet" href="../bower_components/codemirror/lib/codemirror.css">
+  <link rel="import" type="css" href="../bower_components/codemirror/lib/codemirror.css">
+  <template>
+    <style>
+      pre {
+        filter: grayscale(100%);
```
dstockwell: Is this meant to be here?

PR #526, opened 2016-10-17
Read the readme file using the github api, rather than guessing the filename

### LINKED ISSUE #379: Readme isn't processed unless it's cased as README.md
[Origin page](https://beta.webcomponents.org/element/1000ch/lazyload-image)

andymutton:
Fixes #379 for some more common cases.


samuelli:
We should use the GitHub API if we fail to find it.
https://developer.github.com/v3/repos/contents/#get-the-readme

andymutton:
I've completely redone this to just always use the API, since it returns the content of the file in the 'content' field.

This turns the change into a couple of relatively tiny line changes, so I'm happier with it.

src/manage.py
```diff
@@ -587,9 +587,9 @@ def error(self, error_string):
     super(IngestVersion, self).error(error_string)
 
   def update_readme(self):
-    response = urlfetch.fetch(util.content_url(self.owner, self.repo, self.sha, 'README.md'), validate_certificate=True)
+    response = urlfetch.fetch(util.github_url('repos', self.owner, self.repo, 'readme') + '?ref=%s' % self.sha, validate_certificate=True)
```
dstockwell: Need to use github_get here to pass the correct auth tokens.

dstockwell:
Also, you'll probably need to update the tests.

andymutton:
Ch-ch-ch-ch-changes. PTAL.

PR #528, opened 2016-10-17
Use precache option to import runtime cache settings

dstockwell:
Allows the service worker to update when when the embedded checksums change.


dstockwell:
Don't merge this yet.

PR #544, opened 2016-10-19
Send analytics events for preview/publish/toggle-info

dstockwell:


samuelli:
I just sent out a different PR for this

PR #548, opened 2016-10-20
Roll codemirror via npm not bower

samuelli:


dstockwell:
Update client's readme.md?

dstockwell:
Also, why npm?

samuelli:
Bower package is broken. Lib/codemirror.js is missing

PR #550, opened 2016-10-20
Fix stacking issue with code-editor and app-header

samuelli:


client/src/catalog-app.html
```diff
@@ -105,7 +105,7 @@
 
       app-header {
         @apply(--layout-fixed-top);
-        z-index: 1;
+        z-index: 100;
```
dstockwell: hmmm :P

PR #562, opened 2016-10-24
Add yarn.lock files for faster builds 

samuelli:
On cache:
main 70s -> 5s
analysis 20s -> 1s
client 18s -> 1s


dstockwell:
How does the yarn lock file get updated?

samuelli:
It's managed by yarn, so anytime you update, install etc, it'll update the lock file.

PR #572, opened 2016-10-25
Cased repo dependencies & git:// in user content dependencies

samuelli:
Fixes #572

### DIFF
--- a/src/user_content.py
+++ b/src/user_content.py
@@ -37,6 +37,7 @@ class GetResource(webapp2.RequestHandler):
 
     config_map = {}
     for dependency in dependencies:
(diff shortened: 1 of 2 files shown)


src/user_content.py
```diff
@@ -37,6 +37,7 @@ def get(self, owner, repo, tag, name, path):
 
     config_map = {}
     for dependency in dependencies:
+      dependency['owner'] = dependency['owner'].replace('git://github.com/', '')
```
andymutton: Did we decide that this was necessary? Weren't we going to try and fix this in analysis?

samuelli: I thought even after we 'fixed' analysis, we would still output `owner: git://github.com/owner`

PR #582, opened 2016-10-30
Add more fatal error cases. Add tests for them.

andymutton:
Also remove obsolete test that was covering our old parallel version of analysis. It was actually failing but the test wasn't correct.


dstockwell:
Tests are failing because travis has an old version of node :(

PR #605, opened 2016-11-02
Try feature different owners

samuelli:
Since owners tend to update elements in groups, this change attempts to feature different owners by pulling the first 6 and choosing a subset to show.

### DIFF
--- a/client/src/catalog-home.html
+++ b/client/src/catalog-home.html
@@ -161,14 +161,14 @@
       loading="{{_elementsLoading}}"
       url="[[baseUrls.api]]/api/search/kind:element?limit=6&noscore"
       handle-as="json"
-      on-response="_featuredElementsResponse"></iron-ajax>
+      last-response="{{_elementSearch}}"></iron-ajax>
 
     <iron-ajax
       id="featuredCollections"
       loading="{{_collectionsLoading}}"
       url="[[baseUrls.api]]/api/search/kind:collection?limit=6&noscore"
       handle-as="json"
-      on-response="_featuredCollectionsResponse"></iron-ajax>
+      last-response="{{_collectionSearch}}"></iron-ajax>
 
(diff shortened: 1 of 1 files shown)


abdonrd:
Perhaps this work should be done by the API?

samuelli:
@abdonrd: Perhaps, we'll have to add a new `featured` API. ATM, this just uses search which is working out fairly well. Once we hit a performance wall we'll have to switch it over and cache it somehow I think.

client/src/catalog-home.html
```diff
@@ -261,8 +261,30 @@
...
+      },
+
+      _filterFeatured: function(candidates) {
+        var featured = [];
+        var owners = {};
+        // Attempt to feature different owners.
+        for (var i = 0; i < candidates.length && featured.length < 3; i++) {
+          if (!owners[candidates[i].owner]) {
+            owners[candidates[i].owner] = true;
+            featured.push(candidates.splice(i, 1)[0]);
```
dstockwell: seems a bit odd for the filter function to mutate its input, ..and makes the logic complicated. Doesn't this skip over an entry each time you remove a candidate?

samuelli: Rewritten. PTAL.

client/src/catalog-home.html
```diff
@@ -261,8 +261,30 @@
...
+
+      _featuredElementsResponse: function() {
+        this._featuredElements = this._filterFeatured(this.$.featuredElements.lastResponse.results);
+      },
+
+      _featuredCollectionsResponse: function() {
+        this._featuredCollections = this._filterFeatured(this.$.featuredCollections.lastResponse.results);
+      },
+
+      _filterFeatured: function(candidates) {
```
samuelli: Done.

PR #607, opened 2016-11-02
Add manage/analyze-all to trigger re-analysis of all ingested libraries

andymutton:


src/manage.py
```diff
@@ -512,7 +512,7 @@ def handle_get(self, owner, repo):
 
     versions = Version.query(Version.status == Status.ready, ancestor=self.library.key).fetch()
     for version in versions:
-      self.trigger_analysis(version.key.id(), version.sha)
+      self.trigger_analysis(version.key.id(), version.sha, transactional=False)
```
samuelli: optional, or `false`? shrug

andymutton: I was in two minds about it. I decided to just do all of the things.

PR #616, opened 2016-11-04
Fix dependency gathering even more!

andymutton:
Should resolve the demo on - https://beta.webcomponents.org/element/MyScript/myscript-math-web


andymutton:
(Still testing - don't merge)

Issue #648, opened 2016-11-09
Missing dependencies


dstockwell:
https://beta.webcomponents.org/api/docs/Maps4HTML/Web-Map-Custom-Element

Only lists 5 dependencies, but there should be 46 after `bower install`.

andymutton:
With some tweaking, I managed to get one more dependency.

For reasons unknown, bower's info command is not actually returning us the list of devDependencies that we would expect. It's returning us nothing instead. I do not know why, but it does make me weep just a little.

I guess I need to just parse the bower.json or something.

prushforth:
Sorry to make you weep!  I registered the thing with bower so the install should work as 'bower install web-map', but maybe you can't make that assumption when you get a url to a repo.  Let me know if I can tune the repo somehow.  Thanks!

andymutton:
Oh, the install works. I'm just trying to harvest the devDependencies to make sure that the demo works right and I'm using the Bower API to do it. Sometimes it's... obtuse. I have a new plan I'm going to try.

andymutton:
I've figured it out. The error was with me.

@prushforth, the bower.json in your tagged release doesn't contain any devDependencies, which is why the demo doesn't work. I should have looked at what was in my bower_components directory (the difference between master and bower has bitten me more than once - it needs to be my first port of call!).

The correct total number of dependencies for your element as it stands should be 6. I have a fix for the missing one that I'll send out soon.

If you want the demo to work on beta.webcomponents.org, you'll need to make a release that includes the devDependencies in the bower.json.

prushforth:
OK, I did that but I guess you haven't pushed the corrected version yet?

andymutton:
This is true. I just ran it through the staging server. It seems okay so I'll deploy it to production soon.

andymutton:
Okay. Dependency fix is working and dev dependencies are all coming through too now as well.

I just realised that the demo isn't showing up because there isn't an @demo annotation on the web-map element.

I made a fork and tried adding it (https://github.com/andymutton/Web-Map-Custom-Element/blob/master/web-map.html), but that broke the analysis phase, presumably because of problems analysing the demo.

https://custom-elements-staging.appspot.com/element/andymutton/Web-Map-Custom-Element

@prushforth, if you want to work on making it work, there's a https://beta.webcomponents.org/preview URL that you can use to test specific commits and PRs.

I'm going to close this now, since the underlying issue is fixed.

prushforth:
[...]

PR #652, opened 2016-11-14
Don't lowercase git uris when performing the info call

andymutton:
This allows the info call to return info, rather than a cache miss.

We're not _really_ happy with this. It should fix another one of the missing dependency problems, but it's leaning very heavily into the hack world.

Rewriting this code to work differently is starting to seem a sensible approach, although what would be more sensible is probably waiting until we move from hydrolysis to analysis and bower to npm.

### DIFF
--- a/analysis/src/bower.js
+++ b/analysis/src/bower.js
@@ -163,7 +163,7 @@ class Bower {
     return new Promise(resolve => {
       var metadata = null;
       bower.commands.info(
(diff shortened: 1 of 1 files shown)


andymutton:
Fixes #648, partly. In that it adds another dependency. The rest need to be dealt with by publishing the devDependencies in the bower.json for releases as well as in master.

PR #657, opened 2016-11-15
Stop logging speculative failures. Only log on genuine failures.

andymutton:
Adds an array to speculative info calls to collect the errors so that we only log when there are two.

Also just always tries both options, because extra rules probably just cause more bugs.


andymutton:
Travis!! Yarn!!

PR #668, opened 2016-11-21
Fix download info scroll

pdgago:
Use an input so it can be scrollable and call select() on tap.

image

### DIFF
--- a/client/src/catalog-element.html
+++ b/client/src/catalog-element.html
@@ -126,8 +126,14 @@
       }
 
       #info-download-command {
+        border: 0;
+        color: var(--theme-font-color);
+        display: block;
         font-family: var(--monospace-font);
         font-size: 12px;
+        outline: 0;
+        padding: 4px 0;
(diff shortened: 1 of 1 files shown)


client/src/catalog-element.html
```diff
@@ -573,7 +579,7 @@ <h2 class="page-subtitle">[[data.description]] <a title="Home page link" href="[
           <div class="box">
             <div id="info-download">
               <h3>Download</h3>
-              <div id="info-download-command">bower install --save [[owner]]/[[repo]]</div>
+              <input id="info-download-command" value="bower install --save [[owner]]/[[repo]]" on-tap="_selectAllInfoDownload"/>
```
dstockwell: Can this also be readonly? `<input readonly ...`

pdgago: @dstockwell  fixed 👍

samuelli:
Thanks @pdgago!

PR #680, opened 2016-11-29
Show a toast when a new version (new sw registered) is available

samuelli:


dstockwell:
We should add analytics events for display/click on this toast too.

PR #689, opened 2016-12-05
Fix error with field name for dependencies. Add test.

andymutton:


andymutton:
CI error is related to missing java install for wct. Everything else is fine. Submitting.

PR #690, opened 2016-12-05
Preserve analysis data when reanalyzing. Add analysis queue.

andymutton:
Stops us from clearing analysis info when running analyze-all (which would wipe out all data and then reingest slowly).
Adds a dedicated analysis queue to prevent analysis from starving other processes.

### DIFF
--- a/src/manage.py
+++ b/src/manage.py
@@ -312,9 +312,13 @@ class LibraryTask(RequestHandler):
     if self.library.kind == 'collection':
       analysis_sha = sha
     version_key = ndb.Key(Library, self.library.key.id(), Version, tag)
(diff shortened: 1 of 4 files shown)


src/manage.py
```diff
@@ -312,9 +312,13 @@ def trigger_analysis(self, tag, sha, transactional=False):
     if self.library.kind == 'collection':
       analysis_sha = sha
     version_key = ndb.Key(Library, self.library.key.id(), Version, tag)
-    Content(id='analysis', parent=version_key, status=Status.pending).put()
+
+    content = Content.get_by_id('analysis', parent=version_key)
+    if content is None:
```
dstockwell: We should also set status to `pending` if it's currently `error`

PR #693, opened 2016-12-07
Browser compatibility table

### LINKED ISSUE #623: Browser compatibility table to homepage
Design & implement a way to bring the browser compatibility table to the homepage.

samuelli:
Fixes #623 

image
image
image


samuelli:
I think I would agree if the specs were to change, the browser set were to change.

It's only defined once though, so with the links & images it seems more effort than its worth.

PR #705, opened 2016-12-12
Add link to old webcomponents.org

samuelli:


client/src/catalog-app.html
```diff
@@ -487,9 +487,10 @@ <h2 id="domain-title">webcomponents.org</h2>
           <h2>About</h2>
           <div role="navigation">
             <a href="/about">About webcomponents.org</a>
-            <a href="#feedback" on-tap="_onFeedback">Feedback</a>
-            <a href="https://github.com/webcomponents/beta">View on GitHub</a>
-            <a href="/assets">Assets</a>
+            <a href="https://webcomponents.github.io">Deprecated webcomponents.org</a>
```
dstockwell: Deprecated seems a bit weird. Maybe if it was at the end of the list it would be OK?

PR #706, opened 2016-12-12
Clear query on escape while focused inside list

### LINKED ISSUE #617: a11y: When search list is open, <esc> should close it
i.e. when you're focused on an item within the Search dropdown, pressing <Esc> should close the dropdown, remove the query, and refocus the text box.

image

It does this already if the text box is focused, but should do the same thing if a search result is focused as well.

samuelli:
Fixes #617

### DIFF
--- a/client/src/catalog-search-box.html
+++ b/client/src/catalog-search-box.html
@@ -284,6 +284,7 @@
 
         this.addEventListener('focus', focusListener, true);
         this.addEventListener('blur', blurListener, true);
+        this.addEventListener('keydown', this._onKeyDown, true);
       },
 
       open: function() {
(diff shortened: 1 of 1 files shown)


client/src/catalog-search-box.html
```diff
@@ -306,6 +307,14 @@
         this._updateExpandedState();
       },
 
+      _onKeyDown: function(event) {
+        // ESC key
+        if (event.keyCode == 27) {
```
dstockwell: I think you're meant to use `event.code` these days.

samuelli: Nope.

image

PR #732, opened 2016-12-19
Index all in roughly the last updated order

### LINKED ISSUE #729: Maintain ordering when running index-all

samuelli:
Fixes #729


samuelli:
Looks pretty good except for the first few entries: https://custom-elements-staging.appspot.com/elements

Not sure if they don't have the first few don't have updated fields on staging and hence are out of order. vaadin down looks good

Issue #710, opened 2016-12-13
After interaction, component states "undefined"


sbeleidy:
After clicking on one of the buttons in the [demo](https://beta.webcomponents.org/preview/convoo/login-fire/b072f3805f20c0ff2180492d1635365b4307bf7e) for [login-fire](https://github.com/convoo/login-fire/tree/38_fix_demo) the component becomes 'undefined' with no errors displayed in the console.

I had initially assumed the issue was due to data binding but after applying the solution from #701 it's still happening. Previously it wouldn't open the sign-in popups, now it opens them but the element still becomes 'undefined'. You can reproduce by clicking "Sign in with Google" button for example (no need to actually sign in after that).

![screen shot 2016-12-12 at 9 24 05 pm](https://cloud.githubusercontent.com/assets/1388511/21127589/5f87beb4-c0b1-11e6-9ff4-c70937813931.png)

samuelli:
Fixed! It doesn't say `undefined` anymore.

PR #736, opened 2016-12-19
Restrict inline demo post messages to trusted domains

samuelli:
May be related to #710

### DIFF
--- a/inline-demo.html
+++ b/inline-demo.html
@@ -47,7 +47,7 @@
   var demoId;
 
   function receiveMessage(event) {
-    if (!event.origin.match(/https:\/\/\w*\.?webcomponents\.org/))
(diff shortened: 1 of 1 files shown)


inline-demo.html
```diff
@@ -47,6 +47,9 @@
   var demoId;
 
   function receiveMessage(event) {
+    if (!event.origin.match(/^https:\/\/\w*\.?webcomponents\.org/))
```
dstockwell: Can we just test that the data is in the correct format? That is has `id` and `source` properties.

samuelli: Done. Seems not as safe, but I guess it'll work on staging etc now. :\

dstockwell: It's not really a concern in this case. The iframe is already executing on an origin with user content.

PR #756, opened 2017-01-05
Add polymer analyzer to analysis.

andymutton:
This change just adds extra data without any changes to the flow.


andymutton:
(Don't merge - still testing)

andymutton:
I added the API change with parameter as discussed.

PR #798, opened 2017-01-16
Add support for pre-release tags

### LINKED ISSUE #678: Support for semver v1.0.12-1
Could you help me?

Unable to process element - couldn't find any tagged versions

https://beta.webcomponents.org/element/StartPolymer/s-slider
https://github.com/StartPolymer/s-slider

dstockwell:
Fixes #678

### DIFF
--- a/src/api.py
+++ b/src/api.py
@@ -98,6 +98,7 @@ class LibraryMetadata(object):
     }
     if not assume_latest:
       result['latest_version'] = metadata['latest_version']
+      result['default_version'] = metadata['default_version']
     raise ndb.Return(result)
 
   @staticmethod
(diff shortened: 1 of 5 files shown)


src/api.py
```diff
@@ -154,7 +155,7 @@ def full_async(library_key, tag=None, brief=False, assume_latest=False):
       versions = yield versions_future
       result['versions'] = versions
       if len(versions) > 0:
-        result['latest_version'] = versions[-1]
+        result['latest_version'] = versiontag.default_version(versions)
```
samuelli: Should we add `default_version` to the metadata too and add a comment to remove `latest_version`?

src/versiontag.py
```diff
@@ -69,3 +74,19 @@ def categorize(version, existing_versions):
     return 'minor'
 
   return 'patch'
+
+def default_version(versions):
+  """Returns the default version - the latest stable version,
+  or the latests pre-release if all versions are pre-release.
```
samuelli: s/latests/latest/

src/api.py
```diff
@@ -98,6 +98,7 @@ def brief_async(library_key, tag=None, assume_latest=False):
     }
     if not assume_latest:
       result['latest_version'] = metadata['latest_version']
```
samuelli: Doesn't this line need to be updated?

dstockwell: They're both set when we generate the full metadata.

PR #799, opened 2017-01-16
Do not submit: Massaging analyzer data into the hydrolysis form for easier migration 

andymutton:
The plan is to make the new data match the old format by adding the old fields, then always use the new data, then change the elements to use the new data fields, then remove the massaging, then remove the old data.


andymutton:
Weird things with branch...

PR #803, opened 2017-01-17
Revert back to fresh for featured

samuelli:


samuelli:
NB. this only shows 3 elements now not 6

PR #805, opened 2017-01-17
Switch to vanilla progress bar

samuelli:


samuelli:
Works now after an update to the component.

stramel:
@samuelli :+1: you have been kicking some @$$! Thanks!

PR #809, opened 2017-01-18
Don't close the catalog-search-box when meta key is pressed

### LINKED ISSUE #806: In search dropdown, cmd+click opens in new tab but closes search dropdown
I couldn't tell if this was intentional or not. I think the dropdown should remain open so that you can cmd+click multiple results. 

Willing to work on this if the maintainers agree here.

matthewp:
If the meta key is compressed (cmd on OSX, ctrl on Windows), don't close
the catalog-search-box because it means the user has opened the item in
a new tab and might want to continue opening other items.

Closes #806


matthewp:
Looks like I need to rebase. I'll do that.

PR #828, opened 2017-01-25
Transform analyzer data into a more hydrolysis like form for easier migration.

andymutton:
Adds a client-side flag to change the xhr between analyzer and hydrolysis.

Behaviours don't seem to merge all of the inherited fields anymore, so I need to figure out what to do about that.

The migration plan is to make the new data match the old format by adding the old fields, then always use the new data, then change the elements to use the new data fields, then remove the massaging, then remove the old data.


client/src/catalog-element.html
```diff
@@ -722,23 +722,59 @@ <h1 hidden="[[subElement]]">README.md</h1>
...
         return 'editing' in queryParams || prefix == '/preview';
       },
 
       _elements: function() {
         if (!this.docs || !this.docs.content)
           return [];
-        return Object.keys(this.docs.content.elementsByTagName).map(function(elem) {
-          return this.docs.content.elementsByTagName[elem];
+
+        var content = this.docs.content.analyzerData || this.docs.content;
```
samuelli: So is it always going to be `docs.content.analyzerData` moving forward? Would it make more sense if this was `docs.docs` or `docs.analysis` in the API?

andymutton: Changing it to docs.analysis.

Issue #825, opened 2017-01-24
Most pages are missing a title


matthewp:
Except for element listings, most other pages such as community and publish have the same title, `webcomponents.org`.

I'm not too familiar with the "polymer way", would you emit an event from, for example, `<catalog-publish>` called `page-title` or something, and have a parent element set the title that way?

matthewp:
Willing to submit a PR on this if given some advice on the best way to do it.

samuelli:
Event seems fine to me, maybe called `document-title` then the main `<catalog-app>` can listen to that event and append webcomponents.org if it likes. Currently some pages just set it directly - I believe that's what `<catalog-element>` does, but events are probably more 'polymer way'.

PR would be great! Maybe just start with event and 1-2 places of setting the title, then add the rest in a separate PR. How does that sound?

matthewp:
Sounds like a good way to go to me. Any particular pattern you like for titles, for example do you prefer keeping `webcomponents.org` as part of it like:

```
Publish | webcomponents.org
```

Or just

```
Publish element
```

samuelli:
Probably the first, seems to be convention:

Wikipedia has ` - Wikipedia`
Twitter ` | Twitter`
Facebook none
Google ` - Google Search`

matthewp:
Hey @samuelli, it turns out there's not a good place to trigger the event from within the page, because the pages are not inserted each time so you can't used `attached`. They are set to `display: none;` when not visible.

Couple of ideas here, one is to add a page-title attribute to the element, some of the elements already have this. Would be like:

```html
<catalog-community name="community" path="[[_filter('/community', subroute.prefix, subroute.path)]]" page-title="Community"></catalog-community>
```

And then inside of `_pageReady` check for `this.$.pages.selectedItem.getAttribute('page-title')` and use that to set the document.title.

That would be the easiest solution. Another, more robust, solution would be if catalog-app supported a lifecycle callback for when the page become visible. Maybe called `visible`, and if the page implements this callback, catalog-app would call this function after it becomes visible. From there the page could trigger the document-title event.

The benefit of the second idea is that if there was anything else the page wanted to do when it became visible, the callback would allow it to do whatever it needed.

Let me know which you prefer (or if you have another idea).

samuelli:
Hmm, `page-title` is currently used for the top level `<h1>` tag on static pages.

[...]

PR #830, opened 2017-01-25
Change the document's title when a page's visible changes

matthewp:
This is a start of #825, want to get feedback.

The approach is to use the `visible` property that is passed by
iron-pages to each page component, and when that changes to fire a
`document-title` event which is then handled by community-app.

Will update the other pages if this method is accepted.

### DIFF
--- a/client/src/catalog-community.html
+++ b/client/src/catalog-community.html
@@ -260,8 +260,7 @@
         if (!this._content || !this._content.metadata)
           return;
         var metadata = this._content.metadata;
-        // there is no link, so set the title.
-        if(!metadata.link) {
+        if (!metadata.link) {
           this.fire('document-title', metadata.title);
           return;
         }
--- a/client/src/catalog-community.html
+++ b/client/src/catalog-community.html
@@ -156,6 +156,7 @@
           this.$.threshold.clearTriggers();
           this.$.threshold.checkScrollThesholds();
           this._preventManyScrollLoads = true;
+          this.fire('document-title', 'Community');
         }
       },
 
@@ -259,7 +260,7 @@
         if (!this._content || !this._content.metadata || !this._content.metadata.link)
           return;
         window.location = this._content.metadata.link;
-      },
+      }
 
     });
   </script>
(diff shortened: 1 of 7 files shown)


client/src/catalog-community.html
```diff
@@ -259,7 +260,7 @@ <h3 class="page-subtitle">Published on [[_formatDate(_content.metadata.original_
         if (!this._content || !this._content.metadata || !this._content.metadata.link)
```
samuelli: Let's also fire a title change here with the document title.

Some of the /community content is hosted, like the presentations. eg. https://www.webcomponents.org/community/presentations/progressive-performant-polymer-pick-three-at-google-io-2016 so we should update the title there too.

client/src/catalog-static.html
```diff
@@ -80,6 +80,7 @@ <h1 class="page-title">[[pageTitle]]</h1>
 
         window.addEventListener('hashchange', this._hashChangeHandler);
         this.$.ajax.generateRequest();
+        this.fire('document-title', this.pageTitle);
```
samuelli: Maybe `if this.pageTitle`

matthewp:
@samuelli Updated based on your feedback and changes a couple of other pages.

client/src/catalog-community.html
```diff
@@ -256,10 +257,17 @@ <h3 class="page-subtitle">Published on [[_formatDate(_content.metadata.original_
       },
 
       _contentLoaded: function() {
-        if (!this._content || !this._content.metadata || !this._content.metadata.link)
+        if (!this._content || !this._content.metadata)
           return;
+        var metadata = this._content.metadata;
+        // there is no link, so set the title.
```
samuelli: Let's remove this comment, code is pretty readable

client/src/catalog-community.html
```diff
@@ -256,10 +257,17 @@ <h3 class="page-subtitle">Published on [[_formatDate(_content.metadata.original_
...
+        if(!metadata.link) {
```
samuelli: Style nits: Here and elsewhere needs a space after `if`

matthewp:
@samuelli Updated.

PR #833, opened 2017-01-25
Remove 'undefined' element docs

samuelli:
Cases like https://www.webcomponents.org/element/matthewp/wow-tabs & https://www.webcomponents.org/element/PaulKinlan/air-horner


samuelli:
The source would be hydrolysis though?

Issue #840, opened 2017-01-30
Proposal: Fix navigation usability issues on the element page 


samuelli:
Several issues have been filed regarding usability issues on the site:
 * #839 
 * #838 
 * #813 
 * #782 

In summary:
 * Most critically, API docs are hard to find.
 * Vertical space exacerbates this issue, since navigation elements are further down the page.
 * Right-sided navigation - feels odd that buttons on the right affect content on the left. Typical UX means buttons affect content to the bottom and right.

This proposal is for a) moving the sidebar from right to left b) collapsing non-critical information to be dense.

![image](https://cloud.githubusercontent.com/assets/787668/22412624/9691acfc-e704-11e6-8837-d21866fd8dee.png)

Expanded sidebar:
![image](https://cloud.githubusercontent.com/assets/787668/22412628/9e9efb2a-e704-11e6-90aa-a51a67682cd4.png)

samuelli:
Please let me know what you think!
cc/ @tjsavage @dstockwell

tjsavage:
Would the mobile version look the same as today's?

Generally looks reasonable to me. Would love to play with it staged.

Separately - begs the broader question of whether an A/B testing framework would be worthwhile 😄

samuelli:
Mobile version would be the same as desktop now, which is richer than it is currently.
![image](https://cloud.githubusercontent.com/assets/787668/22444093/9d8eebb4-e795-11e6-9fa7-cef9af8ca874.png)


We can do the A/B side, but are there any metrics we're particularly interesting in?

tjsavage:
In this case I think it's probably too early to get meaningful A/B data. But in the longer term I could see having an established way to do this might be useful. But don't want to couple that to this change.

samuelli:
@tjsavage: this is up on [staging](https://custom-elements-staging.appspot.com/element/PolymerElements/marked-element) - use incognito to ensure its not cached by SW.

matthewp:
I like the collapsing non-critical information part of this. the metadata of an element shouldn't be as prominent as it currently is.

But for reasons I can't articulate well, I think having the sidebar on the right is better. Most languages read left to right so having information justified to the left feels more *right* to me. Wish I had a better reason.

samuelli:
@matthewp thanks for the feedback.

Metadata usually flows better in the right. Eg. Npm site. It's less important than the main content.

[...]

PR #842, opened 2017-01-31
Fix navigation issues by moving the sidebar to left, compacting view

samuelli:
Implements #840 
 * Collapsible GitHub info section
 * Compact view of GitHub info
 * Star on GitHub button broken out
 * Moves sidebar from right to left

image
image
image

### DIFF
--- a/client/src/github-info.html
+++ b/client/src/github-info.html
@@ -1,0 +1,251 @@
+<link rel="import" href="../bower_components/polymer/polymer.html">
+
+<dom-module id="github-info">
+  <template>
+    <style>
+      :host {
+        display: flex;
+        flex-direction: column;
+        color: var(--theme-blue-grey);
+      }
+
+      section {
+        padding: 12px 16px;
+      }
+
+      section:not(:first-child) {
+        border-top: 1px solid #ededed;
+      }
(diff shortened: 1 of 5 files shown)


client/src/github-info.html
```diff
@@ -0,0 +1,305 @@
...
+      _s: function(n) {
+        return n == 1 ? '' : 's';
+      },
+
+      _lastUpdated: function(activity) {
+        if (!activity)
+          return;
+        var i = 0;
+        for (i = 0; i < activity.length; i++) {
+          if (activity[activity.length - 1 - i])
```
andymutton: Not that this really matters, but shouldn't this just be a single line for with this line in the exit condition?

for (i = 0; i < activity.length && !activity[activity.length - 1 - i]; i++);

client/src/github-info.html
```diff
@@ -0,0 +1,305 @@
...
+            break;
+        }
+        if (i == 0)
+          return 'Updated recently';
+        if (i == activity.length)
+          return 'Last updated unknown';
```
andymutton: Last update unknown

(with updated, it's a bit awkward)

PR #844, opened 2017-02-01
Use travis' support of yarn

samuelli:
Source: https://blog.travis-ci.com/2016-11-21-travis-ci-now-supports-yarn


.travis.yml
```diff
@@ -26,10 +26,8 @@ env:
 before_install:
 - mkdir -p lib
 install:
-- npm install -g yarn@0.16
-- npm install -g bower
-- npm install -g web-component-tester
 - yarn
+- npm install -g bower
```
abdonrd: Wouldn't it be better to use `yarn global add bower`?

.travis.yml
```diff
@@ -38,10 +36,11 @@ install:
   && printf '\ny\n\ny\ny\n' | ./google-cloud-sdk/install.sh && cd $TRAVIS_BUILD_DIR;
   fi
 - printf '\ny\n' | $HOME/gcloud/google-cloud-sdk/bin/gcloud components install app-engine-python
+- alias wct="$TRAVIS_BUILD_DIR/client/node_modules/web-component-tester/bin/wct"
 script:
 - grunt lint
 - python tests.py $HOME/gcloud/google-cloud-sdk
-- xvfb-run wct --skip-plugin sauce
+- xvfb-run $TRAVIS_BUILD_DIR/client/node_modules/web-component-tester/bin/wct --skip-plugin sauce
 - "(cd analysis; npm test)"
```
abdonrd: Wouldn't it be better to use `"(cd analysis; yarn test)"`?

.travis.yml
samuelli: I'll try it!

.travis.yml
samuelli: I don't see any benefit of yarn as a task runner

abdonrd: I only suggest it to use always the same command.

samuelli: I see. I think people are familiar with the existing one, so I'll leave it for now.

`yarn` build is significantly faster and is deterministic, so its worth the switch here.

PR #866, opened 2017-02-27
Remove elements.polymer-project.org opt out toast

samuelli:


samuelli:
Travis has decided to break our builds.
I've filed: https://github.com/travis-ci/travis-ci/issues/7421

PR #875, opened 2017-03-07
Refactor analyzer code to use updated versions

samuelli:
* Update Google Cloud dependencies to use non-deprecated NPM packages.
 * Use a newer version of polymer-analyzer
 * Refactor how analyzer is called based on polymer-analyzer updates
 * Add a simple test which analyzes a single element from `client/src`


samuelli:
Ping @dstockwell

PR #879, opened 2017-03-08
Add error codes and custom, detailed error messages

### LINKED ISSUE #864: Add error codes
Adding proper error codes will allow for better error messages on the client.

samuelli:
Fixes #864 

Error codes will allow for error messages to be easily changed on the client to be more helpful instead of requiring changes in the database. This is a backwards compatible change and displays error messages with or without codes.

### DIFF
--- a/client/src/catalog-error.html
+++ b/client/src/catalog-error.html
@@ -6,12 +6,24 @@
   <template>
     <style include="catalog-styles"></style>
 
-    <h3>Error [[error.status]]</h3>
+    <h3>Error [[error.status]] - [[_code]]</h3>
     <h1 class="page-title">[[error.title]]</h1>
     <h2 class="page-subtitle">[[error.message]]</h2>
 
     <template is="dom-if" if="[[error.githubPath]]">
       <div>Would you like to view <a href="https://github.com/[[error.githubPath]]">[[error.githubPath]]</a> on Github instead?</div>
+    </template>
+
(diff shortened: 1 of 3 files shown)


client/src/catalog-error.html
```diff
@@ -6,21 +6,61 @@
...
     <h2 class="page-subtitle">[[error.message]]</h2>
 
     <template is="dom-if" if="[[error.githubPath]]">
       <div>Would you like to view <a href="https://github.com/[[error.githubPath]]">[[error.githubPath]]</a> on Github instead?</div>
     </template>
+
+    <template is="dom-if" if="[[_codedError(_code, 5)]]">
+      <h2 class="page-subtitle">Invalid license</h2>
+
+      The GitHub repository must be licensed under an <a href="https://opensource.org/licenses/alphabetical">Open Source Inititative (OSI) approved license</a>. Please include a valid license or alternatively include an OSI approved license as an SPDX identifier (eg. "Apache-2.0") in the license field in the bower.json.
```
samuelli: Leaving this as is because we actually do 2 checks which is a bit inconsistent:
 * First on the [default/master branch](https://github.com/webcomponents/webcomponents.org/blob/master/src/manage.py#L256)
 * On every [version ingestion](https://github.com/webcomponents/webcomponents.org/blob/master/src/manage.py#L651)

PR #886, opened 2017-03-15
Sort elements alphabetically

### LINKED ISSUE #881: Not sorted alphabetically.
Why overview section is not sorted alphabetically. 
untitled

samuelli:
Fixes #881

### DIFF
--- a/client/src/catalog-element.html
+++ b/client/src/catalog-element.html
@@ -496,6 +496,8 @@
             });
           }
           return el;
+        }).sort(function(x, y) {
+          return x.is < y.is ? -1 : 1;
         });
       },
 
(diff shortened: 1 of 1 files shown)


client/src/catalog-element.html
```diff
@@ -496,6 +496,8 @@ <h1 hidden="[[subElement]]">README.md</h1>
             });
           }
           return el;
+        }).sort(function(x, y) {
+          return x.is < y.is ? -1 : 1;
```
dstockwell: I guess they are never the same?

samuelli: Does it matter? It'll just return 1? I guess it might not be a stable sort then?

PR #895, opened 2017-03-17
Add manage API to analyze all latest versions

samuelli:


src/manage_test.py
```diff
@@ -60,6 +60,17 @@ def test_analyze_all(self):
         util.analyze_library_task('owner', 'repo')
     ], [task.url for task in tasks])
 
+  def test_analyze_all_latest(self):
```
dstockwell: should also add to xsrf test above

samuelli: Done

PR #898, opened 2017-03-20
Don't try and scan a package & handle fatal errors

### LINKED ISSUE #897: Resources not loading for tagged release
I've tagged v0.1.0 for my `stock-value` element, but resources don't get loaded in [the webcomponents.org page of the element](https://www.webcomponents.org/element/valdrinkoshi/stock-value)
```
GET https://user-content-dot-custom-elements.appspot.com/valdrinkoshi/stock-value/v0.1.0/stock-value/stock-value.html 404
GET https://user-content-dot-custom-elements.appspot.com/valdrinkoshi/stock-value/v0.1.0/webcomponentsjs/webcomponents-loader.js 404

### PR DESCRIPTION
**Pull request by samuelli**
This fixes #897 as currently it fails to reject the promise when there's an exception. This results in the instance being locked up for 2 minutes and prevents other tasks from being processed.

 * Don't try analyze a package, since we currently don't have the root of the directory. We can probably change this to work in the future.
 * Treat file not found as a fatal error, since this will never recover properly, unless we choose to ignore them in the future. Given this would be 'unintended', maybe failure is better, although it is opaque right now.

### DIFF
--- a/analysis/src/analyzer.js
+++ b/analysis/src/analyzer.js
@@ -32,7 +32,6 @@ class AnalyzerRunner {
         Promise.all(inputs.map((i) => analyzer.analyze(i))).then(function(documents) {
           resolve(generateAnalysis(documents, ''));
         }).catch(function(error) {
-          debugger;
           Ana.fail('analyzer/analyze', inputs, error);
           var fatal = error.code && fatalErrorCodes.indexOf(error.code) != -1;
           reject({retry: !fatal, error: error});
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: analysis/src/analyzer.js [hunk 1]
```diff
@@ -24,16 +27,15 @@ class AnalyzerRunner {
...
-          Ana.fail('analyzer/analyze', inputs, isNotTest);
-        });
+        resolve({});
       } else {
         Promise.all(inputs.map((i) => analyzer.analyze(i))).then(function(documents) {
           resolve(generateAnalysis(documents, ''));
-        }).catch(function() {
-          Ana.fail('analyzer/analyze', inputs);
+        }).catch(function(error) {
+          debugger;
```
dstockwell: oops?

samuelli: Oops.

PR #901, opened 2017-03-22
Add support for bower demos

### LINKED ISSUE #899: Demo support without Hydrolysis / Analyzer [Feature Request]
It seems like right now in order to publish a dedicated demo page with a web component, you have to implement `iron-component-page` and use the inline documentation with the @demo tag. It would be nice to be able to publish a demo page without the dependency on the Polymer hydrolysis/analyzer.

Maybe this could be a custom bower tag or a meta header in the index.html?

samuelli:
Supports demos specified just in `bower.json`. This is an object with titles and paths as key/value pairs respectively. Closes #899.

```
...
"demos": {
    "My Simple Demo": "path/demo.html"
}
...
```

This patch modifies the API to return this specified list and updates the client to dedupe these from any demos that hydrolysis/analyzer might return and display them.

### DIFF
--- a/client/src/catalog-element.html
+++ b/client/src/catalog-element.html
@@ -328,6 +328,10 @@
           <div class="sidebar-nav box loader" role="navigation" collapse$="[[_shouldCollapseNav(bundledElements)]]" loading$="[[!docs]]">
             <a href="[[route.prefix]]/[[owner]]/[[repo]][[versionRoute]]">Overview</a>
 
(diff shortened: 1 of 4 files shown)


samuelli:
[Sample bower.json change](https://github.com/samuelli/progress-bar/commit/51585c9433084fde5f770d015c71006617470748) with how it'll be [shown on the site](https://custom-elements-staging.appspot.com/preview/samuelli/progress-bar/51585c9433084fde5f770d015c71006617470748).

PR #907, opened 2017-03-28
Stable Custom Elements in Safari! 🎉

abdonrd:
screen shot 2017-03-28 at 12 54 46 pm

### DIFF
--- a/client/src/browser-table.html
+++ b/client/src/browser-table.html
@@ -252,10 +252,7 @@
           <div title="Polyfill"></div>
           <div title="Developing"></div>
         </a>
-        <a href="https://webkit.org/blog/7027/introducing-custom-elements/">
-          <div title="Polyfill"></div>
-          <div title="Developing"></div>
-        </a>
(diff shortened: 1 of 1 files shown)


stramel:
Any chance you want to publish this as a separate element? :D

samuelli:
@stramel: The whole table?

stramel:
@samuelli Yeah, would love to embed this into my internal catalog for all my co-workers.

abdonrd:
@samuelli is the deployment to master manual? In Travis I only see the staging.

samuelli:
@stramel: I'll take a look, no guarantees though!

@abdonrd: yep, for two reasons. a) needs to be manually verified before/after deployment and b) since our code is split across 4 different 'services', we need to manually select which ones to deploy.

I've deployed it now; thanks for sending the PR!

samuelli:
@stramel: https://www.webcomponents.org/element/samuelli/wc-browser-table

stramel:
@samuelli Thank you!!! ❤️

Issue #910, opened 2017-03-30
Collection not showing dependencies: simpla-elements


seaneking:
Bug with a particular collection: [simpla-elements](https://www.webcomponents.org/collection/simplaio/simpla-elements) ([repo](https://github.com/simplaio/simpla-elements)). Should show 3 elements in the collection at the moment, but not showing any (though when searching it says it has 3 items - not visible on page or via `dependencies` API).

Am I missing something? You just add the elements as deps to `bower.json` right? Thought it might be a caching issue but last commit was several hours ago.

samuelli:
I think its the `latest`. Best to just use `"simpla-text": "simpla-elements/simpla-text"`.
See: https://github.com/PolymerElements/iron-elements/blob/master/bower.json

seaneking:
Yep that fixed it - why is that the case? `latest` is a supported version name as of Bower 1.2.something

seaneking:
And if this is a known bug should there be a note in the collection publishing page? Can see it catching people out (like me ha)

samuelli:
Probably should be a note! Want to send a PR? Not sure what wording would be best.

https://github.com/webcomponents/webcomponents.org/blob/master/client/src/catalog-publish-collection.html

Probably a bit too minor to fix at this point. Particularly if we change to NPM... one day...

seaneking:
Hopefully one day soon! Just noticed that Bower have effectively put a deprecation warning on their site asking users to move to Yarn. Chomping at the bit to move entirely to Yarn as our recommended workflow for Simpla.

And sure thing will submit a docs PR

samuelli:
Yep, unfortunately you'll have a dependency on bower till all these components are up on NPM. Till then you can add `bower install` to your NPM commands for a 'yarn-like' workflow.

samuelli:
That being said, if you have good insights into using Yarn etc, share with us/community would be great!

seaneking:
Well, we wouldn't be using NPM since wc distributed as HTML imports (as ours are) need a flat dep structure unfortunately. So, Yarn FTW!

Yarn can pull unpublished packages from Github, but there's some oddities around versions (eg: installing Polymer this way doesn't work)

```
yarn add [user]/[element]#^[version] --flat
```

[...]

PR #911, opened 2017-03-30
Add note about 'latest' version bug in collections

seaneking:
From https://github.com/webcomponents/webcomponents.org/issues/910

### DIFF
--- a/client/src/catalog-publish-collection.html
+++ b/client/src/catalog-publish-collection.html
@@ -34,6 +34,8 @@
           <li>(optional) Add an inline demo!</li>
         </ol>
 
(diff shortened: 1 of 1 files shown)


samuelli:
@dstockwell: does version range even work for collections? Or do you have to specify GitHub repo?

samuelli:
Let's restrict it to explicitly owner/repo. The issue isn't really about bower, but we use that owner/repo to look up the owner/repo and create that link. `bower install` doesn't actually get run on collections.

seaneking:
In that case I'd suggest just rephrasing the line about adding bower deps, rather than adding a note (because they're a special kind of bower dep, really). Feel free to close this PR and I can submit a new one if you like

samuelli:
Sounds good.

PR #912, opened 2017-03-30
Support fully qualified URLs in collection bower.json

samuelli:
See https://github.com/cordova-elements/cordova-elements/blob/master/bower.json for example

Causing task queue failure for `task/ensure`.


samuelli:
> test

Added, done.

PR #914, opened 2017-03-31
List of default tags should be proper links

### LINKED ISSUE #905: Header "example searches" aren't really links
The header "links" which perform a search -- "button", "notification", "media", "toolbar", and "image" -- aren't _actually_ links; they're `href="#"`. This means that a middle-button click on them (to open those searches in a new window while you're still reading things on the page you're on) will just open up a new tab pointed at the page you're already looking at, not a page of actual search results.

(If the plan is that they aren't supposed to be links at all and you're supposed to click them and then read results in the popup, and a list of search results of web components is not intended to be a thing that people link to, then maybe they should be `<button>`s or similar so that middle-click doesn't do anything, rather than doing the wrong thing?)

samuelli:
Closes #905

### DIFF
--- a/client/src/catalog-app.html
+++ b/client/src/catalog-app.html
@@ -467,7 +467,7 @@
 
           <div id="tag-list">
             <template is="dom-repeat" items="[[_tagList(7)]]">
-              <a href="/search/[[item]]" on-click="_searchTag">[[item]]</a>
+              <a href$="/search/[[item]]" on-click="_searchTag">[[item]]</a>
             </template>
           </div>
         </div>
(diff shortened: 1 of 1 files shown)


client/src/catalog-app.html
```diff
@@ -467,7 +467,7 @@ <h2 id="domain-title">webcomponents.org</h2>
 
           <div id="tag-list">
             <template is="dom-repeat" items="[[_tagList(7)]]">
-              <a href="#" on-click="_searchTag">[[item]]</a>
+              <a href="/search/[[item]]" on-click="_searchTag">[[item]]</a>
```
TimvdLippe: This should be `href$=`, since `href` is a native attribute

samuelli: Done. I think this doesn't break anyway.

Issue #913, opened 2017-03-31
Support Polymer 2.0


samuelli:
### Required changes

- [x] Run analyzer on all packages #756 
- [x] Run side by side client replacement #923 
- [x] Update client to support analyzer output #932 
- [x] Match existing styling #942 
- [x] Fix performance #942 
- [x] Fix search document indexing to support analyzer #932 
- [x] Use analyzer by default #962 
- [x] Disable hydrolysis for faster build #961 

### External dependencies
- [x] Show inherited/private checkboxes https://github.com/PolymerElements/iron-doc-viewer/issues/123
- [x] Demo list parity https://github.com/Polymer/polymer-analyzer/issues/626

samuelli:
For anyone that might be following along. You can preview Polymer 2.0 by appending `?analyzer` to the URL. Let me know if there's any feedback/bugs.

samuelli:
External dependencies on https://github.com/PolymerElements/iron-doc-viewer/issues/123 and https://github.com/Polymer/polymer-analyzer/issues/626.

PR #923, opened 2017-04-07
Roll analyzer, fix package loading, run side by side

samuelli:
Part of #913 

Summary of key changes:
 * Rolls polymer-analyzer to latest
 * Fixes package loading to use a root path to where it was installed and using relative paths for `mainHtmlPaths`.
 * Package analysis looks through main HTML paths & expands out imported documents while staying within the package.
 * Duplicates `<catalog-element>`, which switches loading to an analyzer supporting version based on the query param `?analyzer`. This means once its ready, we'll switch over directly to the new one without any code debt.


samuelli:
@andymutton: can you look at `analysis.js`? Particularly where I assume that components are installed under `attributes.repo` in the current directory of the process.
@dstockwell: can you look at the rest of the analysis changes? The client/ change is with the duplication & rewrites of `_elements()`, `_behaviors()`. The sidebar populates with elements & behaviors, but otherwise, its not really functioning correctly, but it'll be easier to review once we land the duplicated file, so you can actually see diffs.

client/src/catalog-app.html
```diff
@@ -690,6 +701,7 @@ <h2 id="domain-title">webcomponents.org</h2>
 
         if (isLocalhost)
           result.api = 'https://www.webcomponents.org';
+        result.api = 'https://custom-elements-staging.appspot.com';
```
dstockwell: is this meant to be here?

analysis/src/analysis.js
```diff
@@ -46,9 +47,11 @@ class Analysis {
       this.bower.prune().then(() => {
         return this.bower.install(attributes.owner, attributes.repo, versionOrSha);
       }).then(mainHtmlPaths => {
+        const root = path.resolve(process.cwd(), 'bower_components', attributes.repo);
```
andymutton: Yeah... I wish I could be more helpful here, but I'm not 100% sure that the attributes.repo assumption is correct and I'm also not 100% sure that it's not. My memory on this is fuzzy at best.

Is there a way to verify the correctness with an assert here and then just run a full analysis on staging?

analysis/src/analyzer.js
```diff
@@ -28,9 +28,14 @@ class AnalyzerRunner {
 
       if (inputs == null || inputs.length === 0) {
         resolve({});
+        // TODO: fall back to package analysis
       } else {
         Promise.all(inputs.map((i) => analyzer.analyze(i))).then(function(documents) {
-          resolve(generateAnalysis(documents, ''));
+          var allDocuments = [];
+          documents.forEach(function(doc) {
+            allDocuments = allDocuments.concat(Array.from(doc.getByKind('document', {imported: true, externalPackages: false})));
```
dstockwell: so this still includes all of the main documents?

samuelli: Yes.

Based on offline convos, we were going to ignore main documents & search the entire package, but this turned out to include the wrong things for packages like `app-layout`, which include all sorts of elements for testing, demos, sites etc.

This instead analyses all the main documents, and only includes everything they import while still in the same package.

client/src/catalog-app.html
samuelli: No. Thanks.

analysis/src/analysis.js
samuelli: Sure, I'll do that.

Issue #480, opened 2016-10-12
Add "copy" button for download command on element page


tjsavage:

tjsavage:
& make it scrollable?

samuelli:
Scrollable fixed in #668

bkjohnson:
Can we just leave out the button completely and have the text be copied automatically when you click the bower command?  I'm happy to try this.

samuelli:
Sounds good. Search for `.fire('notify'` on how to trigger a toast.

samuelli:
Added in #940.

PR #940, opened 2017-04-19
Add clipboard copy for #480

bkjohnson:
Copying was straightforward, the `document.execCommand('copy')` command appears to be [supported in modern browsers](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand#Browser_compatibility).

However there is one weird thing about this.  If you click the bower command then it becomes selected, copied, and you see the toast.  If you select something else and then select the bower command *again*, the text again gets copied to your clipboard but you do **not** see the toast a second time.  I verified that the copy was happening by testing with a `concole.log("COPY")` on line 294.  

I also verified that the `notify` event is firing by adding another log in the [handler](https://github.com/webcomponents/webcomponents.org/blob/master/client/src/catalog-app.html#L592) so this seems to be an issue with the toast.  Should I not use the toast at all for this, or try to fix it?

### DIFF
--- a/client/src/catalog-app.html
+++ b/client/src/catalog-app.html
@@ -594,6 +594,7 @@
         this.$.toast.text = options.message;
         this.$.toastAction.textContent = options.buttonText;
         this.$.toast.duration = options.duration == null ? 5000 : options.duration;
+        this.$.toast.removeAttribute('opened'); // In case it was previously set
         this.$.toast.setAttribute('opened', '');
 
         this._currentToast = options;
--- a/client/src/catalog-app.html
+++ b/client/src/catalog-app.html
@@ -593,6 +593,7 @@
         var options = event.detail;
         this.$.toast.text = options.message;
         if (options.hasOwnProperty('buttonText')) {
+          this.$.toastAction.removeAttribute('hidden');
           this.$.toastAction.textContent = options.buttonText;
         } else {
           this.$.toastAction.setAttribute('hidden', '');
--- a/client/src/catalog-app.html
+++ b/client/src/catalog-app.html
@@ -592,7 +592,11 @@
       _notifyUser: function(event) {
         var options = event.detail;
         this.$.toast.text = options.message;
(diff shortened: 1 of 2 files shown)


bkjohnson:
I was able to fix the issue where the toast wouldn't show with this:  https://github.com/bkjohnson/webcomponents.org/commit/711b31acd8e2bed7deb2176c63895512d9dd0d3f

Do you want it in a separate PR?

samuelli:
Method is discouraged since we want to be able to lazily load the toast element. My guess is that the attribute was already on there and its not retriggering open.

Can you add it to the same PR but try using the attributes? eg. removing and adding? If we really can't get it to work, then if the attribute is already set & the open method exists, we can call open() instead.

bkjohnson:
Removing the property worked for me.  Out of curiosity, how does a method call prevent a lazy load?

samuelli:
> Out of curiosity, how does a method call prevent a lazy load?

If you call it before it loads, it'll throw an error (method not found) and won't do the right thing once it does load. Setting the attribute means that it can load after that, see the attribute and knows what to do from there.

client/src/catalog-app.html
```diff
@@ -592,8 +592,14 @@ <h2 id="domain-title">webcomponents.org</h2>
       _notifyUser: function(event) {
         var options = event.detail;
         this.$.toast.text = options.message;
-        this.$.toastAction.textContent = options.buttonText;
+        if (options.hasOwnProperty('buttonText')) {
+          this.$.toastAction.removeAttribute('hidden');
```
samuelli: Is this necessary? Can we just change the next line to `... = options.buttonText || ''`

bkjohnson: The reason I added/removed hidden was because even if the buttonText was empty, the extra space is still going to be there and it looks weird:

image

Would you like me to try an alternative?

samuelli: I see. All good!

PR #942, opened 2017-04-20
Fork iron doc viewer & match existing rendering & styles

samuelli:
Part of #913 

This change automatically forks `iron-doc-viewer#2.0-analyzer` and makes the following changes:
 * Uses existing `mark-down` to render instead of pulling in `marked-element`.
 * Removes `paper-*` dependencies
 * Matches existing styling

Known outstanding work which will be addressed in follow-up PRs to make the deltas much more readable:
 * Fix "Show inherited" & "Show private" check-boxes, which used `paper-checkbox`


stramel:
Just out of curiosity, what is the difference between your mark-down component and the marked-element?

samuelli:
> Just out of curiosity, what is the difference between your mark-down component and the marked-element?

NM. Just that we stripped lots of things to improve performance. The only functional difference I remember is that syntax highlighting in mark-down supports async highlighting, while marked-element is synchronous.  Might try and upstream a fix for that at some point.

stramel:
Yeah that would be awesome especially since we just added remote file support. Thanks for the info!

samuelli:
Updated this pretty significantly FYI @dstockwell. Instead of manually forking, a post install script will copy iron-doc-viewer and perform some rote changes and place the output in `client/src`. This means we'll be able to pull in upstream changes much easier and means its not in our source repo.

PR #950, opened 2017-04-27
Fix API doc error

samuelli:
Regressed in #935, where an error was picked up in the new build, but I fixed it incorrectly.

Considered adding a test, but this is for our hydrolysis docs which I hope to soon no longer use.

Also fixes syntax highlighting in the guessing language case.

### DIFF
--- a/client/src/catalog-api-property.html
+++ b/client/src/catalog-api-property.html
@@ -194,7 +194,7 @@
         <span id="type" class="type">{{descriptor.type}}</span>
         <span class="annotation">[[_getAnnotation(descriptor)]]</span>
       </div>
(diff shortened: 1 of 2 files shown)


client/src/catalog-api-property.html
```diff
@@ -194,7 +194,7 @@
         <span id="type" class="type">{{descriptor.type}}</span>
         <span class="annotation">[[_getAnnotation(descriptor)]]</span>
       </div>
-      <ol id="params" hidden$="{{_computeHideParams(descriptor.return)}}">
+      <ol id="params" hidden$="{{_computeHideParams(descriptor,descriptor.return)}}">
```
dstockwell: space after `,`

PR #951, opened 2017-04-27
Compress analyzer data before storing it

### LINKED ISSUE #938: overview not displaying because "content is too large"
Hey, 

Is there a limit for the size of documentation content that prevents the overview to be displayed in some cases? I have trouble displaying API and demo for two components I wish to register : https://www.webcomponents.org/element/PolymerEl/multi-verse and 
https://www.webcomponents.org/element/PolymerEl/multi-chart?

It seems it tries to load docs from  https://www.webcomponents.org/api/docs/PolymerEl/multi-chart? - which returns http 400 with this error code: "content was too large: 599820".

Thanks,
C.

samuelli:
Fixes #938 

This changes uses gzip encoding to compress data before storing it. For a 700kb analyzer blob, this can be reduced down to ~40kb. In one case, #938, this was > 1mb before hand and ~70kb with the compression.

This doesn't affect the client API since that already uses gzip encoding. Also note that the API only serves hydrolysis or analyzer data and not both.

### DIFF
--- a/src/datamodel.py
+++ b/src/datamodel.py
@@ -2,6 +2,7 @@ from google.appengine.ext import ndb
 
 import re
 import versiontag
+import zlib
 
 class CollectionReference(ndb.Model):
   semver = ndb.StringProperty(indexed=False)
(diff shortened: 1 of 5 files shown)


src/datamodel.py
```diff
@@ -154,12 +155,18 @@ def collections_for_key_async(version_key):
 
 class Content(ndb.Model):
   content = ndb.TextProperty(indexed=False)
+  content_compressed = ndb.BlobProperty(indexed=False)
```
dstockwell: I think BlobProperty actually has an argument to compress automatically, set `compressed=True`

samuelli: PTAL. Using JsonProperty with `compressed=True`.

PR #955, opened 2017-05-04
New raw service in Node

samuelli:
Part of #954.

This duplicates behavior in the user content service into a new 'raw' service. This will allow us to import modules to easily support on the fly transpilation. This service is currently identical to user content with the addition of tests (!).

[Design document](https://docs.google.com/a/google.com/document/d/1sIpX29rR6UQmXtId4tQSk_c_8mbaV9qrmXQmWCUUteE/edit?usp=sharing) for transpilation.

### DIFF
--- a/raw/server.js
+++ b/raw/server.js
@@ -1,8 +1,8 @@
 'use strict';
 
 if (process.env.NODE_ENV === 'production') {
-  require('@google/cloud-trace').start();
-  require('@google/cloud-debug').start();
+  require('@google-cloud/debug-agent').start({allowExpressions: true});
+  require('@google-cloud/trace-agent').start();
 }
 
 const express = require('express');
--- a/raw/server.js
+++ b/raw/server.js
@@ -1,0 +1,56 @@
+'use strict';
+
+if (process.env.NODE_ENV === 'production') {
+  require('@google/cloud-trace').start();
+  require('@google/cloud-debug').start();
+}
+
+const express = require('express');
+const datastore = require('@google-cloud/datastore')();
+const app = express();
+
+// Error on absolute path values.
+app.get('/:owner/:repo/:tag', (request, response) => {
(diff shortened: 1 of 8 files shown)


samuelli:
To test it out, simply point it at the new one after you load any element page.
```javascript
document.querySelector('catalog-element').set('baseUrls.userContent', 'https://raw-dot-custom-elements.appspot.com')
```

raw/server.js
```diff
@@ -0,0 +1,116 @@
...
+  response.set('cache-control', 'max-age=315569000');
+  var url = [req.params.owner, req.params.repo, req.params.tag, req.params.after].join('/');
+  response.redirect(301, url);
+});
+
+app.get('/[^/]+/[^/]+/[^/]+/[^/]+/', (request, response) => {
+  response.sendFile(__dirname + '/inline-demo.html');
+});
+
+app.get('/:owner/:repo/:tag/:name:path(/[\\s\\S]*)', async (req, res) => {
```
dstockwell: why `[\\s\\S]*`, isn't this just `.*`?

raw/server.js
```diff
@@ -0,0 +1,116 @@
...
+  if (!configMap.has(req.params.name)) {
+    res.status(400).send(`${req.params.name} is not a valid dependency for ${tag} in ${owner}/${repo}`);
+    return;
+  }
+
+  // Fetch resource from rawgit
+  const options = {
+    hostname: 'cdn.rawgit.com',
+    path: '/' + configMap.get(req.params.name) + path,
+  }
```
dstockwell: ;

raw/server.js
samuelli: This is a requirement in Express as .* doesn't work. See note at end of [app.param](https://expressjs.com/en/4x/api.html#app.param).

PR #958, opened 2017-05-09
Transpilation on request for raw service

samuelli:
Part of #954.

This change uses the transpilation middleware used in polyserve to a) transpile and b) inject the `custom-elements-es5-adapter` which allows native class-style custom elements to be used after it has been transpiled to es5.

Requests to `raw` are unchanged and transpilation is only when the request is only transformed when the request is prefixed with `/transpile`.

The final piece of the puzzle is to add client side decisions on when to transpile. This can be UA-based determination and/or explicit toggles.


samuelli:
BTW, @dstockwell, the cache control line was taken from user content, but it looks like rawgit sets cache to no-cache. Should we change this to always be cached since URLs are always versioned?

dstockwell:
> BTW, @dstockwell, the cache control line was taken from user content, but it looks like rawgit sets cache to no-cache. Should we change this to always be cached since URLs are always versioned?

Right, I think it's safe to always cache any content responses. Though I don't see the no-cache behavior on rawgit that you mention.

samuelli:
> Right, I think it's safe to always cache any content responses. Though I don't see the no-cache behavior on rawgit that you mention.

My bad, looking at request not response headers. Looks fine on `raw` too.

samuelli:
NB: Added linting & tests to auto run on Travis. Also added caching on Travis for raw and a README.md file too.

PR #967, opened 2017-05-12
Follow redirects and don't set encoding of rawgit responses

### LINKED ISSUE #966: Images are not shown in demo page
Hi,

Since yesterday, I have a problem with images on the demo page, as you can see in the screenshots below and my [custom element](https://github.com/ThomasCybulski/paper-chip). What I'm doing wrong?

Message1

Message2

Thanks!

samuelli:
Fixes #966.

Permanent redirects from rawgit were not being followed and were hence throwing errors. After this was corrected, the byte stream was always interpreted as utf8, which isn't the case for images.

### DIFF
--- a/raw/package.json
+++ b/raw/package.json
@@ -13,7 +13,8 @@
     "@google-cloud/debug-agent": "^1.0.0",
     "@google-cloud/trace-agent": "^1.1.0",
     "express": "^4.15.2",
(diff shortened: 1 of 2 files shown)


samuelli:
Also a related bug that I found in our transpilation service as a result: https://github.com/Polymer/polyserve/pull/195

PR #984, opened 2017-06-07
Support package publishing

samuelli:
Part of #531 

In this change:
 * allows the publish API to accept packages as well as GitHub repos
 * unscoped packages are stored under `@@npm`
 * `update_metadata` calls into the registry to retrieve github repository
 * GitHub repository is stored on the `Library` entity

Next:
 * update version ingestion to pull from package
 * update tag lookup to pull from registry metadata
 * update analysis to pull from package and `npm install`

Issue #531, opened 2016-10-17
Support NPM/yarn (package.json)


lukasoppermann:
I know that Polymer relies on bower, but much of the front-end community has moved entirely to npm. When developing vanilla js custom elements, there is actually no reason to support bower for many developers.

Therefore it would be very good if:

1. one could publish a component with a package that has only a package.json and no bower.json.

2. when an element is available via npm it should be shown on the right as well.

seaneking:
Echoing this sentiment - we want to start preferencing Yarn at [Simpla](https://www.simpla.io), and also move to webcomponents.org for docs/demos of all our elements.

Don't suppose there's any sense of ETA/priority for this one?

seaneking:
Bump. 

Now that Polymer 2 has landed and is [on NPM](https://www.npmjs.com/package/@polymer/polymer) would be great to see webcomponents.org support Yarn out of the box

samuelli:
Up early @seaneking!

I'm thinking of doing pretty minimal support on the webcomponents.org side:
 * You publish a NPM repository, must be linked to a github repo
 * 1-1 mapping between NPM package and GitHub repo
 * Bower is considered supported if bower.json is found, else just NPM/Yarn
 * NPM package gets installed and analyzed for docs & demo serving

WDYT?

seaneking:
:+1: Sounds perfect for an initial release. Thanks @samuelli!

moebiusmania:
@samuelli I agree with @seaneking , this would be a perfect starting point and a real cross-library support! #usetheplatform

moebiusmania:
@samuelli today I have noticed this on the Bower homepage

> ...psst! While Bower is maintained, we recommend yarn and webpack for new front-end projects! 
https://bower.io/

I think that the time has come to start a basic support for NPM registry :)

moebiusmania:
Also, please update the collection publishing process by checking dependencies in ```package.json``` file .

samuelli:
Running list of required technical changes:
- [ ] accept packages #984 
- [ ] update version ingestion to pull from package data
- [ ] update tag lookup to pull from registry metadata
- [ ] update analysis to pull from package and npm install
- [ ] ...more...

moebiusmania:
@samuelli how can we actively help about this? 😊

abdonrd:
Any news?

abdonrd:
Sorry to insist, but, is there any news?

moebiusmania:
[...]

PR #986, opened 2017-06-12
Update version ingestion to pull from package data

samuelli:
Part of #531 

Follows from #984

### DIFF
--- a/src/manage.py
+++ b/src/manage.py
@@ -679,8 +679,9 @@ class IngestVersion(RequestHandler):
     is_npm_package = self.owner.startswith('@')
 
     self.update_readme(is_npm_package)
-    self.update_bower()
-    self.update_pages()
+    if not is_npm_package:
+      self.update_bower()
+      self.update_pages()
     self.set_ready()
 
   def commit(self):
--- a/src/manage.py
+++ b/src/manage.py
@@ -228,14 +228,15 @@ class LibraryTask(RequestHandler):
       self.owner, self.repo = Library.github_from_url(package.get('repository', {}).get('url', ''))
 
       if self.owner == '' or self.repo == '':
-        self.error('No github URL associated with package', ErrorCodes.Library_no_github)
+        return self.error('No github URL associated with package', ErrorCodes.Library_no_github)
 
-        new_metadata = json.loads(response.content)
-        old_metadata = self.library.registry_metadata
-        if old_metadata is not None or new_metadata.get('_rev') != old_metadata.get('_rev'):
-          self.library.registry_metadata = new_metadata
-          self.library.registry_metadata_updated = datetime.datetime.now()
-          self.library_dirty = True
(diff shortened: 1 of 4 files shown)


src/manage.py
```diff
@@ -217,8 +217,7 @@ def pull_registry_info(self):
     # Check if there's a NPM scope or the default scope @@npm.
     assert self.scope.startswith('@')
 
-    headers = {'Accept': 'application/vnd.npm.install-v1+json; q=1.0, application/json; q=0.8, */*'}
-    response = urlfetch.fetch(util.npm_registry_url(self.scope, self.package), headers=headers, validate_certificate=True)
+    response = util.registry_get(self.scope, self.package)
```
dstockwell: this will need to deal with failure, maybe throw an exception before self.retry?

src/manage.py
```diff
@@ -230,6 +229,14 @@ def pull_registry_info(self):
 
       if self.owner == '' or self.repo == '':
         self.error('No github URL associated with package', ErrorCodes.Library_no_github)
+
+        new_metadata = json.loads(response.content)
```
dstockwell: is this meant to be inside the error case?

src/manage.py
dstockwell: is there an etag? how can we avoid fetching this every update, maybe it doesn't matter since there's no quota...

samuelli: There's no etag... it's really quite a terrible, undocumented API. There's no quota, rate limits so I think its ok. I believe the `_rev` will let us know if it has changed so at least we can guard processing/database hits.

src/manage.py
samuelli: No. I'm guessing you wrote this yesterday, its been fixed since then.

src/manage.py
samuelli: Not sure what you mean? The response status code is checked below

dstockwell: Oh, yeah.

src/manage.py
```diff
@@ -217,8 +217,7 @@ def update_registry_info(self):
     # Check if there's a NPM scope or the default scope @@npm.
     assert self.scope.startswith('@')
 
-    headers = {'Accept': 'application/vnd.npm.install-v1+json; q=1.0, application/json; q=0.8, */*'}
-    response = urlfetch.fetch(util.npm_registry_url(self.scope, self.package), headers=headers, validate_certificate=True)
+    response = util.registry_get(self.scope, self.package)
     try:
       package = json.loads(response.content)
```
dstockwell: shouldn't this be inside the `status_code == 200` case?

samuelli: Hmm, those lines aren't really needed. I think we want to assume that we are getting a JSON response and fail loudly if the registry isn't complying with that.

samuelli: Removed

PR #987, opened 2017-06-12
Process tags from registry metadata

samuelli:
Part of #531 

Follows on from #986

### DIFF
--- a/src/manage.py
+++ b/src/manage.py
@@ -468,12 +468,32 @@ class LibraryTask(RequestHandler):
     self.library.tags_updated = datetime.datetime.now()
     return tag_map
 
+  def update_package_tags(self):
+    registry_metadata = json.loads(self.library.registry_metadata)
+    assert registry_metadata is not None
+    versions = registry_metadata.get('versions', {})
+    tags = versions.keys()
+    tags.sort(versiontag.compare)
+    # Create a tag map of tag to sha
+    tag_map = dict((tag, versions.get(tag).get('_shasum', '')) for tag in tags)
+
+    if self.library.tags is None or self.library.tags != tags:
+      self.library.library_dirty = True
+      self.library.tags = tags
+      self.library.tags_map = json.dumps(tag_map)
+      self.library.tags_updated = datetime.datetime.now()
+
+    return tag_map
+
   def update_versions(self):
     if self.library.shallow_ingestion:
       return
 
     if self.library.kind == 'collection':
       new_tag_map = self.update_collection_tags()
+    elif self.scope.startswith('@'):
+      assert self.library.kind == 'element'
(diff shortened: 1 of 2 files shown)


src/manage.py
```diff
@@ -468,12 +468,32 @@ def update_element_tags(self):
...
+    return tag_map
+
   def update_versions(self):
     if self.library.shallow_ingestion:
       return
 
     if self.library.kind == 'collection':
       new_tag_map = self.update_collection_tags()
+    elif self.scope.startswith('@'):
+      assert self.library.kind == 'element'
```
aomarks: Could a user cause this to happen by publishing a weird package.json, or is this just asserting some internal consistency? (I'm wondering if you need a more graceful error to prevent users from killing the pipeline).

src/manage.py
```diff
@@ -468,12 +468,32 @@ def update_element_tags(self):
...
+    versions = registry_metadata.get('versions', {})
+    tags = versions.keys()
+    tags.sort(versiontag.compare)
+    # Create a tag map of tag to sha
+    tag_map = dict((tag, versions.get(tag).get('_shasum', '')) for tag in tags)
+
+    if self.library.tags is None or self.library.tags != tags:
+      self.library.library_dirty = True
+      self.library.tags = tags
+      self.library.tags_map = json.dumps(tag_map)
```
aomarks: Why store as JSON instead of a native dict?

src/manage.py
```diff
@@ -468,12 +468,32 @@ def update_element_tags(self):
     self.library.tags_updated = datetime.datetime.now()
     return tag_map
 
+  def update_package_tags(self):
```
aomarks: Is this part of a pipeline e.g. on a cron job? Or is it part of some synchronous user workflow?

samuelli: Yep. There's a variety of 'tasks' throughout the system. The update tasks get triggered by a cron job every 10 minutes.

src/manage.py
samuelli: Can you do that? What property would that map to? https://cloud.google.com/appengine/docs/standard/python/datastore/typesandpropertyclasses

Regardless, this also matches existing behavior.

aomarks: Oh cool, actually I didn't even realize this was a datastore object.

src/manage.py
samuelli: In the normal flow, we have two 'kinds'. Element repos and collection repos. They're both GitHub repos with collections defining the 'element-collection' keyword in the bower.json. This then gets processed differently.

At the moment, all I've implemented is some support for element packages. I'm not really sure how collections will be structured yet.

This check is basically checking internal consistency since we don't process packages as collections yet (and perhaps never will..)

PR #994, opened 2017-06-20
Update analysis service & other fixes

samuelli:
* Update analysis dependencies and move away from deprecated vm to flex environment
 * Don't validate certificate for registry... because it appears to be invalid
 * Misc fixes

### DIFF
--- a/analysis/package.json
+++ b/analysis/package.json
@@ -18,7 +18,7 @@
     "stream-buffers": "3.0.1"
   },
   "engines": {
-    "node": "7"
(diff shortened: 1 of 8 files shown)


analysis/package.json
```diff
@@ -18,7 +18,7 @@
     "stream-buffers": "3.0.1"
   },
   "engines": {
-    "node": "6"
+    "node": "7"
```
aomarks: why not 8?

samuelli: Done

Issue #928, opened 2017-04-12
Markdown ## headers being stripped from custom documentation page


chrismbeckett:
This is a weird one, but I have updated my markdown document a few times to verify. If you are using the new custom pages feature, and the markdown file has a `##` header line, it gets removed during rendering.

UPDATE: If I change the `##` and use HTML `<h2>` then it works. Seems like it might strip `#` headers as well.

samuelli:
Hmmm, yeah.

We should probably fix this in a better way. Because of a lot of readmes duplicate the title of the repo in it, I wanted to hide it in those cases. So this is the source in question, which doesn't work well for guides.

```
article.markdown-body > h2:first-of-type, article.markdown-body > h1:first-of-type {
        display: none;
      }
```

chrismbeckett:
Ah, I just came back to update that using h2 doesn't solve it. It seemed to work because I put two headers in a row. Based on your info above, that makes sense - it only hides the first one.

Ok, well, easy enough to work-around for now, I will just divide up my overview to add an extraneous header so that the doc looks ok whether it display or not. Thanks for the info.

I see the wisdom in hiding the first h1, since most README's just about always start with the repo name. Why hide the first h2? In my README's (following the sample for publishing), it is usually the `## Installation` block that ends up having the header trimmed.

seaneking:
I noticed this too (see https://github.com/webcomponents/webcomponents.org/issues/988). Echoing @chrismbeckett, what's the rational for hiding h2 first-of-type? I would hazard a guess that the use case of using h2s in a README for subheadings is far bigger than using a h2 for the main readme title.

seaneking:
That said, more robust solution would be to do a simple regex check for h1/h2 that contains element name with JS. In the meantime though I think a quick fix for the majority case would be to just remove the h2 decl. Feels pretty buggy for most pages right now

samuelli:
I've been pretty swamped working on other things, so if someone wants to take a look and send a PR that'll be greatly appreciated :)

Otherwise, I'll take a look whenever I can.

seaneking:
👍 Happy for a PR to just ignore h2's for now?

samuelli:
Sure. Or the regex you described above if its the exact same name.

seaneking:
https://github.com/webcomponents/webcomponents.org/pull/997

seaneking:
@samuelli Example of the regression #1000 caused (originally flagged in #997 ):

[...]

PR #997, opened 2017-06-21
Patch greedy subtitle hiding

seaneking:
Addresses https://github.com/webcomponents/webcomponents.org/issues/928

Was going to do a proper JS fix, but don't have time to setup the buildstep for local testing atm.

@samuelli logic would have looked something like this, if you wanted to throw it in yourself (somewhere around [this line](https://github.com/webcomponents/webcomponents.org/blob/master/client/src/catalog-element-readme.html#L233))

```js
var headings = [].slice.call(Polymer.dom(this.$.contents).querySelectorAll('h1,h2'));
var elementNameExp = this.data.repo.split('-').join('[ -]');
var headingsRegex = new RegExp(elementNameExp, 'i');
        
headings.find(function(element) {
  element.innerText.match(headingsRegex);
}).style.display = 'none';
```

### DIFF
--- a/client/src/catalog-element-readme.html
+++ b/client/src/catalog-element-readme.html
@@ -22,7 +22,7 @@
         color: inherit;
       }
 
-      article.markdown-body > h2:first-of-type, article.markdown-body > h1:first-of-type {
+      article.markdown-body > h1:first-of-type {
         display: none;
       }
 


samuelli:
Thanks for this! I modified your code a little into #1000.

seaneking:
@samuelli #1000 wouldn't cover the case where the title is something like `[Component] [Name]` (rather than `component-name`) would it? That's a very common pattern IMO. Why not a simple regex search? It's already pretty narrow, since you're only targeting the first occurrence of h1/h2.

PR #1000, opened 2017-06-23
Conditionally hide first heading in readme.md

### LINKED ISSUE #928: Markdown ## headers being stripped from custom documentation page

samuelli:
Fixes #928

### DIFF
--- a/client/src/catalog-element-readme.html
+++ b/client/src/catalog-element-readme.html
@@ -22,10 +22,6 @@
         color: inherit;
       }
 
-      article.markdown-body > h2:first-of-type, article.markdown-body > h1:first-of-type {
-        display: none;
-      }
-
       :host[shorten] article.markdown-body > :not(:nth-child(-n+4)) {
         display: none;
       }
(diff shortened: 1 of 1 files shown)


client/src/catalog-element-readme.html
```diff
@@ -230,6 +226,11 @@
         links = Polymer.dom(this.$.contents).querySelectorAll('[src]');
         for (i = 0; i < links.length; i++)
           links[i].setAttribute('src', links[i].getAttribute('src').replace(relativeRegex, githubBase + '/raw/' + defaultBranch + '/$1'));
+
+        // Hide first heading if its the same as the repo name.
+        var firstHeading = this.$.contents.querySelector('h1, h2');
+        if (firstHeading && firstHeading.textContent.match(this.data.repo))
```
aomarks: Should this just be `includes` or `indexOf` instead of `match`? I assume `this.data.repo` is not intended to be a regexp.

samuelli: Probably - changed.

PR #1004, opened 2017-06-27
Fix hash URLs in readmes

### LINKED ISSUE #992: README TOCs broken (linkify headings)
Would be really great to have the same auto-linkifying of headings in the pulled in READMEs as GitHub does, because right now GH-style TOCs in READMEs (or other md docs) are broken on webcomponents.org element pages

Examples:

- https://www.webcomponents.org/element/SimplaElements/simpla-article

- https://www.webcomponents.org/element/SimplaElements/simpla-collection

samuelli:
Fixes #992

### DIFF
--- a/client/src/catalog-element-readme.html
+++ b/client/src/catalog-element-readme.html
@@ -200,7 +200,7 @@
        * Github's rendering of the markdown.
        */
       _hashChanged: function() {
-        var hash = window.location.hash.replace('#', '');
+        var hash = window.location.hash.substring(1);
         var element = this.$.contents.querySelector('a#user-content-' + hash);
         if (element)
           element.scrollIntoView();
--- a/client/src/catalog-element-readme.html
+++ b/client/src/catalog-element-readme.html
@@ -1,4 +1,5 @@
 <link rel="import" href="../bower_components/polymer/polymer.html">
+<link rel="import" href="../bower_components/iron-location/iron-location.html">
 
 <dom-module id="catalog-element-readme">
   <!-- TODO: Support for external stylesheets is deprecated in favor of style modules. -->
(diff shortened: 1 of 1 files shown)


client/src/catalog-element-readme.html
```diff
@@ -193,6 +195,17 @@
         '_loadDemos(docsPending)',
       ],
 
+      /**
+       * This handles the hashchange event and finds the element that corresponds based on
+       * Github's rendering of the markdown.
+       */
+      _hashChanged: function() {
+        var hash = window.location.hash.replace('#', '');
```
aomarks: substring(1) because simpler and also the hash is allowed to have its own hash (I do that in iron-doc-viewer!)

client/src/catalog-element-readme.html
```diff
@@ -193,6 +195,17 @@
...
+        var element = this.$.contents.querySelector('a#user-content-' + hash);
```
aomarks: `this.$.contents.$$` or `Polymer.dom(this.$.contents.root)`?

client/src/catalog-element-readme.html
```diff
@@ -231,6 +244,10 @@
         var firstHeading = this.$.contents.querySelector('h1, h2');
         if (firstHeading && firstHeading.textContent.indexOf(this.data.repo) != -1)
           firstHeading.style.display = 'none';
+
+        // Loaded with a hash in URL.
+        if (window.location.hash)
```
aomarks: why wouldn't the iron-location listener do this?

client/src/catalog-element-readme.html
samuelli: It's not a custom element or shadow root, just a regular old element.

client/src/catalog-element-readme.html
samuelli: Content isn't there yet since its loaded async

PR #1009, opened 2017-06-28
Update API to return both package and repo information

samuelli:
Part of #531

### DIFF
--- a/API.md
+++ b/API.md
@@ -4,7 +4,7 @@ Before using the API, please file an issue with your intention of using the API.
 You can access the API at `www.webcomponents.org/api/<method>`. Example search API call: https://www.webcomponents.org/api/search/input.
 
 ## NPM packages
-For relevant endpoints, owner and repo can be replaced my the NPM's scope and package
+For relevant endpoints, owner and repo can be replaced by the NPM's scope and package
 name. For packages without a scope, specify the scope `@@npm`.
 
 ## Get data for an author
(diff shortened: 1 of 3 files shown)


API.md
```diff
@@ -3,6 +3,10 @@ Before using the API, please file an issue with your intention of using the API.
 
 You can access the API at `www.webcomponents.org/api/<method>`. Example search API call: https://www.webcomponents.org/api/search/input.
 
+## NPM packages
+For relevant endpoints, owner and repo can be replaced my the NPM's scope and package
```
aomarks: "replaced by the"

src/api.py
```diff
@@ -130,6 +130,13 @@ def full_async(library_key, tag=None, brief=False, assume_latest=False):
       raise ndb.Return(None)
 
     result = {}
+    # Add NPM package fields
+    key = library_key.string_id()
+    if key.startswith('@'):
+      if key.split('/')[0] != '@@npm':
```
aomarks: double equals?

aomarks: nit: make a parts variable instead of splitting 3 times?

API.md
samuelli: Done

src/api.py
samuelli: Python not js

samuelli: Done.

PR #1011, opened 2017-06-29
Hide more cases of page titles

samuelli:
Fixes the case reported in #997

### DIFF
--- a/client/src/catalog-element-readme.html
+++ b/client/src/catalog-element-readme.html
@@ -248,7 +248,7 @@
         // Treat space and - as interchangeable
         headingContent = headingContent.replace(/[ -]/, '[ -]');
         var headingRegex = new RegExp(headingContent, 'i');
-        if (this.data.repo.match(headingRegex) || (this.pageTitle && this.pageTitle.match(headingRegex)))
+        if (firstHeading && (this.data.repo.match(headingRegex) || (this.pageTitle && this.pageTitle.match(headingRegex))))
           firstHeading.style.display = 'none';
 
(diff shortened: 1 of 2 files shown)


client/src/catalog-element-readme.html
```diff
@@ -244,8 +244,11 @@
 
         // Hide first heading if its the same as the repo name.
         var firstHeading = this.$.contents.querySelector('h1, h2');
-        var headingContent = firstHeading && firstHeading.textContent.replace(/\s*<?([^>\s]*)>?\s*/, '$1');
-        if (headingContent == this.data.repo || (this.pageTitle && headingContent == this.pageTitle))
+        var headingContent = firstHeading && firstHeading.textContent.replace(/\s*<?([^>]*)>?\s*/, '$1') || '';
+        // Treat space and - as interchangeable
+        headingContent = headingContent.replace(/[ -]/, '[ -]');
+        var headingRegex = new RegExp(headingContent, 'i');
+        if (this.data.repo.match(headingRegex) || (this.pageTitle && this.pageTitle.match(headingRegex)))
```
aomarks: Not blocking, but to avoid accidental matches from treating the name as a regexp, maybe a regular compare after normalization (e.g. `toLowerCase().replace(/[ -]/g, '')` is better.

samuelli: I thought about this. But because it's comparing multiple values I would have to normalise all of them first. eg. `toLowerCase()` and strip `[ -]` on all of them so I thought the regex would be simpler.

PR #1015, opened 2017-07-01
Client UI Fix: Add padding/margin for author sidebar in large view 

### LINKED ISSUE #1014: Client UI Bug: padding/margin for author sidebar + content
Thanks for the great work!

Noticed a possible UI bug when > 800px (width) for the Author Sidebar / Content areas.

The lack of spacing between the sidebar (flex order 2) and the content area is particularly noticeable when the author only has elements listed, as opposed to also having collections.

webcomponents_org

Issue is still noticeable here - when the author has collections (cards).

image

I'll be throwing a PR up for a fix!

tsaiDavid:
Fixes #1014

I chose to follow the established padding/margin used in the media query styles for affected containers.

cursor_and_tsaidavid

### DIFF
--- a/client/src/catalog-styles.html
+++ b/client/src/catalog-styles.html
@@ -89,11 +89,14 @@
         flex: 10 0 270px;
         max-width: calc(100% - 270px);
         min-width: 300px;
+        padding: 0 16px;
       }
 
       .sidebar {
         flex: 0 0 270px;
         order: 2;
+        margin: 16px;
+        margin-top: 0;
       }
 
       .sidebar h3 {


client/src/catalog-styles.html
```diff
@@ -89,11 +89,14 @@
         flex: 10 0 270px;
         max-width: calc(100% - 270px);
         min-width: 300px;
+        padding: 0 16px;
```
samuelli: Can we remove this? It affects multiple pages and affects alignment with the top nav.

client/src/catalog-styles.html
```diff
@@ -89,11 +89,14 @@
...
       }
 
       .sidebar {
         flex: 0 0 270px;
         order: 2;
+        margin: 16px;
+        margin-top: 0;
```
samuelli: Let's change this to margin-left: 40px

PR #1026, opened 2017-07-07
Update license on update tasks

### LINKED ISSUE #1023: License doesn't refresh if changed
Noticed that after changing the license on github for an element registered to webcomponents site it doesn't change to reflect. Is there a way to trigger a cache rebuild of an element or to check this after registering? It's not super common to change licenses but it would be good if they were in sync.

samuelli:
Fixes #1023.

The `update_license_and_kind()` function was not called during the update cycle and only during the initial ingestion task.


samuelli:
@dstockwell: why was this excluded in the first place?

PR #1035, opened 2017-07-19
Fix element counts

### LINKED ISSUE #1032: Counter of elements is incorrect
[Origin page](https://www.webcomponents.org/elements)

samuelli:
Fixes #1032

### DIFF
--- a/src/api.py
+++ b/src/api.py
@@ -34,6 +34,7 @@ class SearchContents(webapp2.RequestHandler):
     index = search.Index('repo')
     cursor = search.Cursor(web_safe_string=request_cursor)
     try:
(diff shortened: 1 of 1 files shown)


src/api.py
```diff
@@ -34,7 +34,7 @@ def get(self, terms):
     index = search.Index('repo')
     cursor = search.Cursor(web_safe_string=request_cursor)
     try:
-      accuracy = 100 if include_count else None
+      accuracy = 2000 if include_count else None
```
aomarks: Maybe add a quick note about what accuracy means, since it's unintuitive.

PR #1043, opened 2017-07-31
Add promise-polyfill. Fixes #1027.

### LINKED ISSUE #1027: webcomponents.org broken in IE11
Hi,

Sadly the site seems to be not working in IE11... Going to the top of the site and clicking on 'Browse Elements' just gives me an empty page. No errors in the console.

Thanks,
Martin.

arthurevans:
For simplicity, added the polyfill to the main app element rather than adding it individually to each element that uses iron-ajax.

### DIFF
--- a/client/src/catalog-element.html
+++ b/client/src/catalog-element.html
@@ -1089,7 +1089,7 @@
 
       _absolutePath: function(url) {
         // String.startsWith not supported on IE11
-        return (url.substr(0,4) === 'http');
+        return url.substr(0, 4) === 'http';
       },
 
       _docPageLoaded: function(event) {
--- a/client/src/catalog-element.html
+++ b/client/src/catalog-element.html
@@ -1088,7 +1088,8 @@
       },
 
       _absolutePath: function(url) {
-        return url.startsWith('http');
+        // String.startsWith not supported on IE11
+        return (url.substr(0,4) === 'http');
       },
 
       _docPageLoaded: function(event) {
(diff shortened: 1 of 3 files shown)


arthurevans:
R: @samuelli 

This seems to fix the error in the client, although I wasn't able to test with all of the services using my local VM.

samuelli:
Thanks @arthurevans!

I've deployed it here: https://custom-elements-staging.appspot.com/

Can you verify it works on IE11?

arthurevans:
Hi Sam. Verified that this fix gets IE working better, but demos weren't showing up because IE doesn't have `String.startsWith`. It appears that this is only used one place in the client code, so I hacked around it in place instead of trying to polyfill it.

Verified that this change doesn't break other browsers.

client/src/catalog-element.html
```diff
@@ -1088,7 +1088,8 @@ <h1 hidden="[[subElement]]">[[_pageTitle(_customPages, pagePath)]]</h1>
       },
 
       _absolutePath: function(url) {
-        return url.startsWith('http');
+        // String.startsWith not supported on IE11
+        return (url.substr(0,4) === 'http');
```
samuelli: needs a space `(0, 4)` and the outer parentheses aren't necessary

arthurevans:
Thanks! Fixed.

samuelli:
FYI, this has been deployed to webcomponents.org

arthurevans:
Verified I can now see the docs and run the demo for iron-ajax using IE11. Thanks, @samuelli!

PR #1047, opened 2017-08-07
Upgrade code to Polymer 1.x/2.x hybrid support

FluorescentHallucinogen:
See https://www.polymer-project.org/2.0/docs/upgrade#convert-template-extension-elements-at-the-document-level.


FluorescentHallucinogen:
@samuelli PTAL.

PR #1058, opened 2017-09-06
Add script type=module to support matrix

samuelli:
image

### DIFF
--- a/client/src/browser-table.html
+++ b/client/src/browser-table.html
@@ -106,13 +106,13 @@
         flex-direction: column;
       }
 
-      [title="Stable"]:before, [title="Polyfill"]:before {
+      [type="Stable"]:before, [title="Stable"]:before, [title="Polyfill"]:before {
         background: url("data:image/svg+xml;utf8,<svg xmlns=%27http://www.w3.org/2000/svg%27 viewBox=%270 0 22 22%27 version=%271.1%27 aria-hidden=%27true%27><path fill-rule=%27evenodd%27 fill=%27white%27 d=%27M12 5l-8 8-4-4 1.5-1.5L4 10l6.5-6.5z%27></path></svg>"), #8edb69;
         background-position: 5px 3px;
       }
 
 
-      [title="Developing"]:before {
+      [type="Developing"]:before, [title="Developing"]:before {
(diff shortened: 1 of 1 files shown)


samuelli:
Changed Imports to HTML imports

client/src/browser-table.html
```diff
@@ -263,7 +263,15 @@ <h1 label>Browser support</h1>
         </a>
       </div>
       <div class="row">
-        <div label="Imports"><svg viewBox="0 0 26 22"><use xlink:href="/sprite.octicons.svg#import"></use></svg></div>
+        <div label='<script type="module">'><svg viewBox="0 0 26 22"><use xlink:href="/sprite.octicons.svg#import"></use></svg></div>
```
abdonrd: Why not `ES modules` o something similar?
There are not `<template>` or `<link rel="import">` in the others.

samuelli: Because there's multiple parts that are separately tracked. eg. [dynamic import](https://www.chromestatus.com/feature/5684934484164608) is listed separately. We'll probably include that later (its hard to find info on it and its not very interesting to show in the table yet).

Eventually we'll replace this with one row of "ES Modules".

samuelli: Does that make sense? @abdonrd

stramel: Hmm wonder if the row of ES Modules with the row containing support for the parts? Modules, nomodule, and dynamic imports?

samuelli: Hmm, if we did that, it would have to be its own table with 3 rows.

matthewp:
Note that Firefox is *not* developing shadow DOM or custom elements. Here's the bug where it became unassigned for [shadow DOM](https://bugzilla.mozilla.org/show_bug.cgi?id=1205323#a58584152_442218) and for [custom elements](https://bugzilla.mozilla.org/show_bug.cgi?id=889230).  Should I submit a new PR to fix this?

client/src/browser-table.html
matthewp: There's also worker modules. I created https://aremodulesready.com for this purpose :)

abdonrd: @samuelli got it! Makes sense.

@stramel @matthewp thanks for the feedback!

PR #1075, opened 2017-11-16
Change export markdown to download

### LINKED ISSUE #1074: Markdown button in editing mode not working
The markdown button in editing mode (live demo editor) seems to be broken. 

> Not allowed to navigate top frame to data URL

[see](https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/GbVcuwg_QjM)

samuelli:
Fixes #1074.

Opening data URIs in a top level frame has been disabled in Chrome.

### DIFF
--- a/client/src/custom-element-demo.html
+++ b/client/src/custom-element-demo.html
@@ -297,12 +297,7 @@
           '```',
         ].join('\n');
 
-        // Initiate a download of the markdown. Opening data URIs in the top frame
-        // was deprecated.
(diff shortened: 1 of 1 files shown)


samuelli:
> Looks like it's also possible to work around it using a blob, if you would rather continue to show it in a new window.

how?

dstockwell:
> how?

```
window.open(URL.createObjectURL(new Blob(['hello'], {type: 'text/plain'})))
```

PR #1086, opened 2017-12-10
Add links to corresponding accepted specifications

TimvdLippe:
I am always googling my way to the correct location of the specifications. Seems like a nice fit for this table to also provide a link to the correct specification location.

### DIFF
--- a/client/src/browser-table.html
+++ b/client/src/browser-table.html
@@ -227,7 +227,9 @@
         </div>
       </div>
       <div class="row">
-        <div label="Templates"><svg viewBox="0 0 26 22"><use xlink:href="/sprite.octicons.svg#template"></use></svg></div>
+        <a href="https://html.spec.whatwg.org/multipage/scripting.html#the-template-element" target="__blank" rel="noopener">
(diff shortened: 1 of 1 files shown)


abdonrd:
Should we update the links?

I think we can use always the latest editor's drafts, since it has the most up-to-date information. Also, they have the links to the latest published versions.

The latest editor's drafts:
- https://w3c.github.io/webcomponents/spec/custom/
- https://w3c.github.io/webcomponents/spec/shadow/ (waiting https://github.com/w3c/webcomponents/issues/661)
- https://w3c.github.io/webcomponents/spec/imports/

The latest published versions:
- https://www.w3.org/TR/custom-elements/
- https://www.w3.org/TR/shadow-dom/
- https://www.w3.org/TR/html-imports/

TimvdLippe:
I am okay with that as well. Don't really mind to which they link 😄

samuelli:
Either seems fine to me!

abdonrd:
PR at https://github.com/webcomponents/webcomponents.org/pull/1097

PR #1087, opened 2017-12-11
Fix styling in browser table

samuelli:
Followup after styling broke in #1086

### DIFF
--- a/client/src/browser-table.html
+++ b/client/src/browser-table.html
@@ -53,20 +53,25 @@
 
       /* Spec label */
 
-      div[label] {
+      [spec] {
         flex: 0 0 170px;
+        display: flex;
         margin-right: 16px;
+        color: var(--theme-blue-small);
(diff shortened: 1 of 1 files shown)


client/src/browser-table.html
```diff
@@ -269,7 +273,7 @@ <h1 label>Browser support</h1>
         </a>
       </div>
       <div class="row">
-        <a href="https://html.spec.whatwg.org/multipage/webappapis.html#integration-with-the-javascript-module-system" target="__blank" rel="noopener">
+        <a spec href="https://html.spec.whatwg.org/multipag/webappapis.html#integration-with-the-javascript-module-system" target="_blank" rel="noopener">
```
TimvdLippe: The `multipage` was correct. It results in a 404 now 😢

samuelli: My bad!

PR #1090, opened 2017-12-23
Fix breakpoints when page is ~600px wide

### LINKED ISSUE #1089: Responsive Layout Can't Break Properly on set width
[Image](https://polymer.slack.com/files/U74AV5PRB/F8GQH6146/webcomp.png)

Responsive layout does not hide/show items in order. Also doesn't fire at set responsive width (I think it's supposed to be around 580px). I have seen this problem on this website and my own Polymer Projects. But I don't think the problem is with app-layout, I think the problem comes from "Layers" being all over the place. I have checked many Polymer Projects, they all have layers, all over the place.

samuelli:
Fixes #1089

### DIFF
--- a/client/src/catalog-element.html
+++ b/client/src/catalog-element.html
@@ -250,18 +250,20 @@
         color: hsl(0, 0%, 40%);
       }
 
-      @media (max-width: 500px) {
-        .title {
-          margin-left: 16px;
-          display: flex;
-        }
-
+      @media (max-width: 700px) {
         .sidebar {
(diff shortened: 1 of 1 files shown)


solutionboxing:
This does not fix the problem. You set the break point to 700px, But the .sidebar element has a margin-right of 25px. Therefore the app breaks at 675px. See .sidebar {margin: -32px 25px -64px -15px;} <-- margin information never gets passed down for the app to work properly.

Issue #1098, opened 2018-01-05
Flag text in Browser Table


Link2Twenty:
In the Browser Table the test reading "FLAG IN 54" is just entered in the HTML meaning it doesn't get removed when the page gets smaller (or on mobile).

The problem is [browser-table#L282](https://github.com/webcomponents/webcomponents.org/blob/eccd72b4515b5a086c99612c1b2638fb1d5d4a17/client/src/browser-table.html#L282)
```HTML
<a type="Developing" title href="https://developer.mozilla.org/en-US/Firefox/Experimental_features#JavaScript">Flag in 54</a>
```

![flag-text](https://user-images.githubusercontent.com/3534427/34611801-dbe66c36-f21e-11e7-83cd-bc3bb56d7670.PNG)

I'd suggesting doing something like;

```HTML
<a href="https://developer.mozilla.org/en-US/Firefox/Experimental_features#JavaScript">
  <div title="Flag" data-version="54"></div>
</a>
```

```CSS
[type="Flag"]:before, [title="Flag"]:before {
  background: url("data:image/svg+xml;utf8,<svg xmlns=%27http://www.w3.org/2000/svg%27 viewBox=%270 0 22 22%27 version=%271.1%27 aria-hidden=%27true%27><circle fill=%27white%27 cx=%2750%%27 cy=%2750%%27 r=%272%27></circle></svg>"), #f5a623;
 }
[type="Flag"], [title="Flag"]:after {
  color: hsla(37,91%,34%,1);
  content: 'FLAG IN ' attr(data-version);
}
```

The new title/type would need to be added to [browser-table#L159](https://github.com/webcomponents/webcomponents.org/blob/eccd72b4515b5a086c99612c1b2638fb1d5d4a17/client/src/browser-table.html#L159) too.

Link2Twenty:
Come to think of it this might be more elegant

```HTML
<a title="Developing" data-version="54" href="https://developer.mozilla.org/en-US/Firefox/Experimental_features#JavaScript" class="style-scope browser-table"></a>
```
```CSS
[data-version]:after {
  content: 'FLAG IN ' attr(data-version);
}
```
Just need to add `data-version` and it will change Developing to FLAG IN 54 and it should treat the text the same a the normal Developing text.

PR #1099, opened 2018-01-08
Add data-version for FLAG IN comment

Link2Twenty:
Manually inputting the flag text lead to the table looking messy when viewport width was small.
Now you can add data-version="[VER]" to any of the marks to change the text to "FLAG IN [VER]"
This new text will behave the same as all text and vanish when the screen is small enough.

See issue #1098


samuelli:
Looks good, thanks!

abdonrd:
FYI: https://twitter.com/FirefoxNightly/status/951382754125545473

`<script type="module">` enabled in Nightly builds 🎉

Issue #1095, opened 2018-01-02
Header is horizontally misaligned in IE11


limonte:
![](https://i.imgur.com/WWXAYiq.png)

Link2Twenty:
This line is causing the issue [catalog-app#L89](https://github.com/webcomponents/webcomponents.org/blob/8a24bff7c7ecd38678ade199c21e6b0924e26374/client/src/catalog-app.html#L89).

Removing it centres the header, I tested removing it in chrome too and it had no issues.

Link2Twenty:
Just realised this is a repeat of issue #835

limonte:
Yup, it's a duplicate of #835 which hasn't been fixed for almost 1 year.

Issue #835, opened 2017-01-26
Header not centered in IE11


web-padawan:
See this screenshot:

![ie11](https://cloud.githubusercontent.com/assets/10589913/22344094/5ddb9598-e403-11e6-957f-64fee1c1edbb.png)

samuelli:
Thanks for the report - I don't have access to IE11 rn, would you like to send a PR?

PR #1102, opened 2018-01-12
Removed margin: 0 auto;

Link2Twenty:
This line seems redundant in all browsers (Chrome, Firefox and Edge) but breaks alignment in IE.
Removing it fixes the alignment for IE and changes nothing in the other browsers

See issues #835 and #1095


samuelli:
Can you confirm this fixes the linked issues and close them once the changes are live? @Link2Twenty

Issue #1065, opened 2017-10-11
audio/video from demo page keeps playing after closing the demo dialog


valdrinkoshi:
[Origin page](https://www.webcomponents.org/element/mpachnis/mp-video)

To repro:
- open an element's demo page with audio/video, e.g. https://www.webcomponents.org/element/mpachnis/mp-video/demo/demo/index.html
- play the video
- close the dialog

Notice the video is still playing, as the iframe is still in the dom

Link2Twenty:
A quick fix for this would be to clear the iframe's source when the dialog is closed, as it generated the iframe afresh anyway.

```JS
document.querySelector("catalog-dialog#demo-dialog > iframe").src=""
```

Link2Twenty:
This shouldn't be an issue anymore.

PR #1104, opened 2018-01-12
Empty iframe src on _demoCanceled

Link2Twenty:
When closing a demo dialog the iframe within it is still active meaning actions, audio playback/recording etc, will continue after the demo is 'closed'. When a demo is open the src is set regardless of its current value, causing the frame to refresh, as this is the case there is no need to keep the src once the demo is closed.

See issue #1065


samuelli:
Awesome, thanks for the pull request!

PR #1121, opened 2018-02-15
Ignore minified, bundled files in analysis

samuelli:


analysis/src/analyzer.js
```diff
@@ -26,7 +26,25 @@ class UrlLoader extends FSUrlLoader {
...
    */
   async readDirectory(pathFromRoot, deep) {
-    return await new FSUrlLoader(this.packageRoot).readDirectory(pathFromRoot, deep);
+    const files = await new FSUrlLoader(this.packageRoot).readDirectory(pathFromRoot, deep);
+    const result = [];
+
+    // Filter out minified files etc.
+    for (const path of files) {
+      // eg. for file.min.js => file.js
+      const fileMatch = pathlib.basename(path).match(/([^.]*)(?:\.[^.]*)*(\.[^.]*)/);
```
aomarks: Kinda hard to read this regexp. Would probably just prefer using `split` or something. Up to you though.

samuelli: Thanks, changed to use `split`.

PR #1123, opened 2018-02-21
Copy node_modules folder before analyzing

samuelli:
This change allows node_modules folder to be analyzed since it is currently stripped about by the analyzer.


analysis/src/npm.js
```diff
@@ -63,7 +65,7 @@ class NPM {
     return new Promise((resolve, reject) => {
       Ana.log("npm/prune");
       fs.writeFileSync('installed/package.json', '{}');
-      childProcess.exec("rm -rf installed/node_modules", function(err) {
+      childProcess.exec("rm -rf installed/node_modules installed/modules_copy", function(err) {
```
aomarks: why not use a node library to delete?

I like fs-extra, probably can replace ncp too.

samuelli: Not sure, I didn't write the original. Changed to use fs-extra.

analysis/src/npm.js
```diff
@@ -60,18 +58,17 @@ class NPM {
...
-      childProcess.exec("rm -rf installed/node_modules", function(err) {
-        if (err) {
-          Ana.fail("npm/prune");
-          reject({retry: true, error: err});
-        } else {
-          Ana.success("npm/prune");
-          resolve();
-        }
-      });
+      fs.writeJsonSync('installed/package.json', {});
```
aomarks: await

analysis/src/npm.js
```diff
@@ -60,18 +58,17 @@ class NPM {
   * @return {Promise} A promise handling the prune operation.
   */
   prune() {
-    return new Promise((resolve, reject) => {
+    return new Promise(async (resolve, reject) => {
```
aomarks: just make this whole function async/await instead of making a promise yourself?

samuelli: Not sure how that works with the reject params, sorry.

PR #1124, opened 2018-02-22
Filter analysis result

samuelli:
Filtering files for analysis was not sufficient. If a file depends on a file that was not found in the search, it can still end up in the analyzed output. This change filters results before final output of analyzing results. Filtering before is still useful as that forms analyzer's search roots and can help prevent some additional overhead associated with having to analyze additional files.

### DIFF
--- a/analysis/src/analyzer.js
+++ b/analysis/src/analyzer.js
@@ -69,10 +69,12 @@ class AnalyzerRunner {
         analyzer.analyzePackage().then(analysis => {
           // Get source paths for all features.
           const features = Array.from(analysis.getFeatures());
-          const paths = features.map((f) => f.sourceRange && f.sourceRange.file);
+          const paths = features
+            .filter((f) => f.sourceRange !== undefined)
+            .map((f) => f.sourceRange.file);
           const uniquePaths = new Set(paths);
 
-          const shouldOuputFeature = (feature) => {
+          const shouldOutputFeature = (feature) => {
             if (!isInPackage(feature)) {
               return false;
             }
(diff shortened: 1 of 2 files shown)


analysis/src/analyzer.js
```diff
@@ -67,7 +67,31 @@ class AnalyzerRunner {
 
       if (paths.length === 0) {
         analyzer.analyzePackage().then(analysis => {
-          resolve(generateAnalysis(analysis, analyzer.urlResolver, isInPackage));
+          // Get source paths for all features.
+          const features = Array.from(analysis.getFeatures());
+          const paths = features.map((f) => f.sourceRange && f.sourceRange.file);
```
aomarks: this would map to undefined if sourceRange was undefined, so really you should filter. I guess if you're sure it can't be undefined, just use `!`

analysis/src/analyzer.js
```diff
@@ -67,7 +67,31 @@ class AnalyzerRunner {
...
+          const uniquePaths = new Set(paths);
+
+          const shouldOuputFeature = (feature) => {
```
aomarks: Output

analysis/src/analyzer.js
```diff
@@ -67,7 +67,31 @@ class AnalyzerRunner {
...
+
+            const path = feature.sourceRange.file;
+            const fileMatch = pathlib.basename(path).split('.');
+            if (fileMatch.length <= 2) {
+              return true;
+            }
+
+            const strippedBaseName = fileMatch[0] + '.' + fileMatch[fileMatch.length - 1];
+            // Don't use pathlib.join() as that will strip file:/// protocol.
+            if (!uniquePaths.has(pathlib.dirname(path) + '/' + strippedBaseName)) {
```
aomarks: use path.sep then?

analysis/src/analyzer.js
samuelli: Done.

analysis/src/analyzer.js
samuelli: Done - not TS so `!` not needed.

analysis/src/analyzer.js
samuelli: Done.

PR #1131, opened 2018-03-26
Center the main page footer

ExE-Boss:
Right now, the footer is left aligned, this PR instead centers it, so that it feels more aesthetically balanced.

### DIFF
--- a/client/src/catalog-app.html
+++ b/client/src/catalog-app.html
@@ -385,6 +385,7 @@
         color: white;
         padding: 32px 0;
         flex-wrap: wrap;
+        margin: 0 auto;
       }
 
       #footer-content section {


Link2Twenty:
Ah, that's my fault I took it out of a multi-selector and didn't put it back in with a separate `#footer-content` selector.
https://github.com/webcomponents/webcomponents.org/commit/e884e64e43bda52fc890dbadff05de363e5c8748#diff-8701d227be581f8526bac69ef329ffc4.

We could use flexbox to solve this too if we'd rather.

```CSS
footer {
  display: flex;
  justify-content: center;
}
```

https://github.com/webcomponents/webcomponents.org/blob/master/client/src/catalog-app.html#L377-L381

Might be overkill though.

ExE-Boss:
I think using flexbox would be overkill too, and using a flexbox instead of margins might also have some unintended side-effects.

Issue #1127, opened 2018-03-06
Bower NPM interop


samuelli:
- [x] Bower page auto redirects to NPM page
- [x] NPM page has a link back to the Bower page
- [x] Bower continues to update but bower.json must exist it accepts the new version
- [x] Bower version won't show up in search if its also on NPM

PR #1134, opened 2018-03-26
Migrate instead of suppressing bower -> npm

samuelli:
Part of #1127

### DIFF
--- a/client/src/catalog-element.html
+++ b/client/src/catalog-element.html
@@ -386,7 +386,7 @@
 
           <div id="github-action-group" class="box">
             <template is="dom-if" if="[[data.npmPackage]]">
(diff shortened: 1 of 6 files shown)


aomarks:
I dont think I understand the high level description "Migrate instead of suppressing bower -> npm" can you clarify

samuelli:
> I dont think I understand the high level description "Migrate instead of suppressing bower -> npm" can you clarify

Before, when a package was moved from bower to NPM, we would suppress the Bower version. This means it was not deleted, but it won't be visible anymore. Per #1127, we'd like this to be more transitional - bower docs are accessible still but there's better interop between the two. This changes suppression to a notion of 'migrated' so it still functions but we can explicitly check fields as to where/to it was migrated from.

PR #1146, opened 2018-05-02
Filter out test files and remove dupes

samuelli:


analysis/src/analyzer.js
```diff
@@ -61,6 +61,11 @@ class AnalyzerRunner {
       const rootUrl = await analyzer.urlResolver.resolve('');
       const isInPackage = feature => feature.sourceRange &&
         feature.sourceRange.file.startsWith(rootUrl);
+      // Filter out test files. This matches both `/test/abc.html` and
+      // `/my-test.html`.
+      const testRegex = /\btests?\b/;
```
aomarks: This seems pretty aggressive -- any thing with the word test in it at all. E.g. this would break the test-fixture element I assume. Maybe just filtering out a top-level test (and demo?) directory is enough?

client/src/catalog-element.html
```diff
@@ -908,6 +909,13 @@ <h1 hidden="[[subElement]]">[[_pageTitle(_customPages, pagePath)]]</h1>
               });
             }
 
+            // Descriptor object can have multiple instances of the same
```
aomarks: They shouldn't.

analysis/src/analyzer.js
samuelli: Done. PTAL @aomarks

analysis/src/analyzer.js
```diff
@@ -61,6 +61,17 @@ class AnalyzerRunner {
       const rootUrl = await analyzer.urlResolver.resolve('');
       const isInPackage = feature => feature.sourceRange &&
         feature.sourceRange.file.startsWith(rootUrl);
+      // Filter out top level test/ and demo/ directories.
+      const isTest = (feature) => {
+        if (!feature.sourceRange) {
+          return false;
+        }
+        const relativePath = feature.sourceRange.file.substring(rootUrl.length);
+        if (relativePath.startsWith('test/') || relativePath.startsWith('demo/')) {
```
aomarks: maybe "tests?/"

samuelli: Not a regex, but I added a check for `tests/`.

aomarks: lgtm

PR #1148, opened 2018-05-04
Fix link to NPM package when displayed in card styles

samuelli:


client/src/collection-card.html
```diff
@@ -153,6 +153,12 @@ <h2>[[data.repo]]</h2>
         return version == latestVersion ? '' : '/' + version;
       },
 
+      _packagePath: function(data) {
+        if (data.npmFullPackage)
+          return data.kind + '/' + data.npmFullPackage;
```
aomarks: why not use `.join()` in both places

samuelli: Fixed

PR #1149, opened 2018-05-09
Handle multi-package repositories

samuelli:
In cases where a mono repo is used to publish many packages there are a few assumptions that are no longer true:
 - Versions of packages are not necessarily associated with a particular SHA. NPM registry will no longer return a relevant `gitHead` SHA that can be used.
 - Pulling the GitHub repository readme is no longer correct, since each package typically has its own readme.

This change fixes these two assumptions:
 - It removes the expectation of a SHA for a given NPM version
 - It pulls the readme by looking at the NPM registry metadata for the readme filename, then pulling the readme file contents via unpkg.

### DIFF
--- a/src/manage.py
+++ b/src/manage.py
@@ -747,32 +747,34 @@ class IngestVersion(RequestHandler):
     super(IngestVersion, self).error(error_string)
 
   def update_readme(self, is_npm_package):
-    github_owner = self.owner
-    github_repo = self.repo
-
     if is_npm_package:
-      # Load parent library object to get the repo information
+      # Load registry metadata to fetch readme path.
       library = Library.get_by_id(Library.id(self.owner, self.repo))
-      github_owner = library.github_owner
-      github_repo = library.github_repo
+      registry_metadata = json.loads(library.registry_metadata) if library.registry_metadata else None
+      readmePath = registry_metadata.get('readmeFilename', 'README.md')
+      response = util.unpkg_get(self.owner, self.repo, self.version, readmePath)
+      readme = response.content
+      print readme
+    else:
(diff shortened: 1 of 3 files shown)


src/manage.py
```diff
@@ -747,32 +747,34 @@ def error(self, error_string, code=None):
...
-      github_owner = library.github_owner
-      github_repo = library.github_repo
-
-    print github_owner, github_repo
-    response = util.github_get('repos', github_owner, github_repo, 'readme', params={"ref": self.sha})
+      registry_metadata = json.loads(library.registry_metadata) if library.registry_metadata else None
+      readme_path = registry_metadata.get('readmeFilename', 'README.md')
+      response = util.unpkg_get(self.owner, self.repo, self.version, readme_path)
+      readme = response.content
+      print readme
```
aomarks: did you really want to print?

src/manage.py
```diff
@@ -747,32 +747,34 @@ def error(self, error_string, code=None):
...
-      readme = base64.b64decode(json.loads(response.content)['content'])
+      if response.status_code == 200:
+        readme = base64.b64decode(json.loads(response.content)['content'])
+      elif response.status_code == 404:
+        readme = None
+      else:
+        return self.retry('error fetching readme (%d)' % response.status_code)
 
+    if readme is not None:
+      # Store the raw readme markdwon contents.
```
aomarks: markdown

src/manage.py
samuelli: no

src/manage.py
samuelli: fixed.

PR #1152, opened 2018-05-09
Fix spec names

FluorescentHallucinogen:


client/src/browser-table.html
```diff
@@ -289,7 +289,7 @@ <h1 label>Browser support</h1>
       </div>
       <div class="row">
         <a spec href="https://w3c.github.io/webcomponents/spec/imports/" target="_blank" rel="noopener">
-          <div label="HTML imports"><svg viewBox="0 0 26 22"><use xlink:href="/sprite.octicons.svg#import"></use></svg></div>
+          <div label="HTML Imports"><svg viewBox="0 0 26 22"><use xlink:href="/sprite.octicons.svg#import"></use></svg></div>
```
samuelli: I wonder if we should simply remove this HTML imports from this table. WDYT?

client/src/browser-table.html
```diff
@@ -247,7 +247,7 @@ <h1 label>Browser support</h1>
       </div>
       <div class="row">
         <a spec href="https://w3c.github.io/webcomponents/spec/custom/" target="_blank" rel="noopener">
```
ExE-Boss: This is now located at <code>https://html.spec.whatwg.org/multipage/custom-elements.html</code>

client/src/browser-table.html
FluorescentHallucinogen: No-no-no, we in no case should not delete HTML Imports. Almost all existing web components (at least published on https://webcomponents.org) still use HTML Imports. ES modules support lands in all browsers and Polymer 3.0 released just yesterday. :)

FluorescentHallucinogen: @ExE-Boss Could you please provide some proof that HTML Imports spec has been actually deprecated? :-/ I know, that Firefox and Safari teams has no plans to implement HTML Imports support, while the current status of HTML Imports in Edge is ["Under consideration"](https://developer.microsoft.com/en-us/microsoft-edge/platform/status/htmlimports/). But I have not yet heard that it's officially deprecated. What about "Not recomended"? ;-)

ExE-Boss: Yeah, <q>Not recommended</q> is probably better.

client/src/browser-table.html
FluorescentHallucinogen: @ExE-Boss Should we use https://www.w3.org/TR/html-imports/ instead of https://w3c.github.io/webcomponents/spec/imports/ for HTML Imports?

client/src/browser-table.html
```diff
@@ -279,7 +279,7 @@ <h1 label>Browser support</h1>
       </div>
       <div class="row">
         <a spec href="https://html.spec.whatwg.org/multipage/webappapis.html#integration-with-the-javascript-module-system" target="_blank" rel="noopener">
-          <div label='<script type="module">'><svg viewBox="0 0 26 22"><use xlink:href="/sprite.octicons.svg#import"></use></svg></div>
+          <div label='ES Modules'><svg viewBox="0 0 26 22"><use xlink:href="/sprite.octicons.svg#import"></use></svg></div>
```
ExE-Boss: I’ve never actually seen these be called <q>ES Modules</q>, only <q>ES6 Modules</q>, also the label attribute should use `"` instead of `'`.

FluorescentHallucinogen: @ExE-Boss See e.g. https://hacks.mozilla.org/2018/03/es-modules-a-cartoon-deep-dive/ and https://medium.com/web-on-the-edge/es-modules-in-node-today-32cff914e4b. See also https://bytearcher.com/articles/es6-vs-es2015-name/, https://benmccormick.org/2015/09/14/es5-es6-es2016-es-next-whats-going-on-with-javascript-versioning/ and https://codeburst.io/javascript-wtf-is-es6-es8-es-2017-ecmascript-dca859e4821c for more info. :wink:

client/src/browser-table.html
samuelli: Latter seems best.

PR #1171, opened 2018-05-22
Add `alt` text attributes to the badges in the readme

ExE-Boss:
Some of the badges in the Readme take a while to load, or fail to load at all (especially the build status and staging server badges), this adds the `alt` text attributes to them, so that there’s fallback text in place while the image loads.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,10 +3,10 @@
   <img alt="webcomponents.org" src="https://web-components-resources.appspot.com/static/logo.svg" width="161">
 </p>
 <p align="center">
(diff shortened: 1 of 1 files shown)


Link2Twenty:
It looks like they still appear blank, even with the alt tags

https://github.com/ExE-Boss/webcomponents.org/blob/31b736da58948362bac50253ff58fba0611cfd97/README.md

ExE-Boss:
It works on my machine (maybe because I use Firefox):

WebComponents.org – Working alt attributes

PR #1174, opened 2018-05-29
Fix for Results page

Link2Twenty:
Problem with `total-results`
Was using camelCase in the HTML

Also added `totalResults` to `_title` to listen for changes.


samuelli:
Thanks for the PR. When I tried this PR out, it didn't work. Seems like `totalResults` never gets set on the parent and hence the `title` function never gets executed. Following up on the bug in #1176

PR #1201, opened 2018-08-21
Update bower since registry changed URLs

### LINKED ISSUE #1188: analysis fails
I just published an element to webcomponents but notice a spinning wheel on the page 

element page - https://www.webcomponents.org/element/epiviz/data-filter

The call to https://www.webcomponents.org/api/docs/epiviz/data-filter?use_analyzer_data returns a 400. 

issue #1042 mentions maybe this is an issue with `bower install epiviz/data-filter`, but I ran this on a clean directory and see no errors.

any ideas why this is failing?

samuelli:
Fixes #1188.

Bower registry changed its URL, which is causing bower tasks to fail.


aomarks:
Looks like a legitimate test failure

samuelli:
> Looks like a legitimate test failure

PTAL @aomarks. Updated to just update bower instead of touching the NPM part.

PR #1202, opened 2018-08-22
WIP Promote package to be analyzed and use PackageUrlResolver

samuelli:


analysis/src/analysis.js
```diff
@@ -48,15 +49,18 @@ class Analysis {
...
         return this.bower.install(attributes.owner, attributes.repo, versionOrSha);
-      }).then(result => {
+      }).then(async (result) => {
         if (!fs.existsSync(result.root)) {
           Ana.fail("analysis/processNextTask", taskAsString, "Installed package not found");
           reject({retry: false, error: Error("Installed package not found")});
           return;
         }
 
+        // Move /bower_components/element/* to /bower_components/*.
```
aomarks: shouldn't this be "Move /bower_components/element* to /*"?

aomarks: Add short explanation like: It's simpler to analyze a layout where the package we are interested in is at the root with its dependencies in bower_components/, than when the package is a sibling of its dependencies.

analysis/src/analysis.js
```diff
@@ -84,11 +88,20 @@ class Analysis {
 
       this.npm.prune().then(() => {
         return this.npm.install(attributes.owner, attributes.repo, attributes.version);
-      }).then(root => {
+      }).then(async(root) => {
```
aomarks: I'm a little confused about what `root` is here. What does `npm.install` return?

analysis/src/analysis.js
```diff
@@ -84,11 +88,20 @@ class Analysis {
...
+        let newRoot = root;
+        // Find the node_modules folder. This may be nested since it could be a
+        // scoped package.
+        while (path.basename(newRoot) !== 'node_modules') {
+          newRoot = path.dirname(newRoot);
+        }
+        // Move up to parent of node_modules.
+        newRoot = path.dirname(newRoot);
+        // Move /node_modules/element/* to /node_modules/*.
```
aomarks: to /*

analysis/src/analysis.js
```diff
@@ -84,11 +88,20 @@ class Analysis {
...
+        await hoistPackageContents(root, newRoot);
+
         return Promise.all([
```
aomarks: await Promise.all

analysis/src/analyzer.js
```diff
@@ -2,65 +2,30 @@
...
-}
-
 class AnalyzerRunner {
   analyze(root, inputs) {
     return new Promise(async (resolve, reject) => {
       Ana.log('analyzer/analyze', inputs);
 
       var paths = inputs || [];
 
+      const isBower = paths.length > 0;
```
aomarks: how does this mean it's bower? (I don't know what inputs or paths are)

analysis/src/analysis.js
```diff
@@ -48,15 +49,18 @@ class Analysis {
       var versionOrSha = attributes.sha ? attributes.sha : attributes.version;
       this.bower.prune().then(() => {
         return this.bower.install(attributes.owner, attributes.repo, versionOrSha);
-      }).then(result => {
+      }).then(async (result) => {
         if (!fs.existsSync(result.root)) {
```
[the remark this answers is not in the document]
samuelli: Done

analysis/src/analysis.js
samuelli: Yes! Thanks

samuelli: Done

analysis/src/analysis.js
samuelli: It returns the location of the installed package.

samuelli: Renamed variables and added a comment here.

analysis/src/analysis.js
samuelli: Thanks.

analysis/src/analysis.js
samuelli: Done

analysis/src/analyzer.js
samuelli: Sorry, this is an implementation detail that's passed in from another place. In bower we have the "main" field, so we pass in the list of input files. In NPM we don't, so we analyze the whole package.

I've switched this around so this is now injected into this function, and this function operates differently depending on whether or not we said it was bower.

PR #1210, opened 2018-09-12
Rewrite sibling node_module references

samuelli:


demo/src/html-rewriter.ts
```diff
@@ -105,10 +105,18 @@ export class HTMLRewriter extends RewritingStream {
...
-        if (attribute && attribute.value === 'module') {
+        const typeAttribute = startTag.attrs.find(({name}) => name === 'type');
+        if (typeAttribute && typeAttribute.value === 'module') {
           insideModuleScript = true;
         }
+
+        // Rewrite any sibling references to node_modules folder as absolute
+        // paths.
+        const srcAttribute = startTag.attrs.find(({name}) => name === 'src');
+        if (srcAttribute && srcAttribute.value.startsWith('../node_modules')) {
```
aomarks: need to also support e.g. `../../node_modules` if the demo was in a subdirectory. maybe something using path.resolve can do it.

PR #1257, opened 2019-09-11
Update to node 10 std env

arthurevans:


arthurevans:
/cc @aomarks 

Staged updated demo service at npm-demos-test:

https://npm-demos-test.appspot.com/@polymer/paper-icon-button@3.0.2/demo/index.html

Staged a hacked version of the client at https://node-10-demos-dot-custom-elements.appspot.com that uses the npm-demos-test instead of the official npm-demos URL. See that working here:

https://node-10-demos-dot-custom-elements.appspot.com/element/@polymer/paper-icon-button/demo/demo/index.html

Because of the way most of these webcomponents docs are set up, you can't just browse to the page for paper-icon-button and click "Demo"; that takes you back to the official webcomponents.org. So you need to munge up the URLs yourself to test. 

For a super-simple change, I found it quite confusing trying to set things up so I could verify that the services were working correctly together. So I added some notes which hopefully help the next bewildered traveller. (But probably won't, because they'll have some completely different problem. So it goes.)

PR #1320, opened 2022-09-02
Clear out the repo for a new implementation

justinfagnani:
Follow-on PRs will set up the first packages and common infrastructure.


.github/CODEOWNERS
```diff
@@ -1 +1,2 @@
 * @arthurevans
+* @justinfagnani
```
aomarks, line 2: Do we actually want codeowners? I'm not sure it's actually too useful. It means you and arthur will get assigned to PR.

aomarks (approved):
LGTM but wondering about codeowners

.github/CODEOWNERS
justinfagnani: We've been doing it for Lit repos, and this was already here... I thought it might be mildly useful in a monorepo when we might have different owners for different folders. I don't care so much.

PR #1323, opened 2022-09-13
Add catalog-server and catalog-api packages

justinfagnani:
`catalog-api` contains the GraphQL schema for the forthcoming GraphQL API to the catalog. It's useful here, before implementing the GraphQL endpoint, to define the schema interfaces that the catalog implementation will be dealing with.

`catalog-server` is a package that will implement the GraphQL endpoint. This PR add a `Catalog` class with a couple of basic operations: import package version, and get package version. A `Catalog` relies on two other interfaces `Repository` and `PackageFiles. `Repository` represents the database, and there's currently only one implementation for Firestore. `PackageFiles` is the interface for retrieving package metadata and files, usually from the npm registry and a CDN like unpkg.com. There is one implementation for production, and test implementations that load form memory or local files.


packages/catalog-api/src/lib/schema.graphql
```diff
@@ -0,0 +1,237 @@
...
+"""
+Information about a specific version of a package
+"""
+type PackageVersion {
+  version: String!
+
+  """
+  The dist tags assigned to this version
+  """
+  distTags: [String!]!
```
rictic, line 130: what happens when we notice that a distTag is moved to a new version? we write to both PackageVersions?

packages/catalog-api/src/lib/schema.graphql
```diff
@@ -0,0 +1,237 @@
...
+
+  author: String!
+
+  tagName: String
+
+  className: String
+
+  """
+  Reference to the custom element export
```
rictic, line 216: What's the format for a reference?

packages/catalog-server/src/lib/firestore/custom-element-converter.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import {
+  DocumentData,
+  FirestoreDataConverter,
+  QueryDocumentSnapshot,
+} from '@google-cloud/firestore';
+
+import {CustomElement} from '@webcomponents/catalog-api/lib/schema.js';
+
+export const customElementConverter: FirestoreDataConverter<CustomElement> = {
```
rictic, line 15: Oh the converters run client side, not in Firebase itself? That makes sense I suppose

packages/catalog-server/src/lib/npm.ts
```diff
@@ -0,0 +1,58 @@
...
+export class NpmAndUnpkgFiles implements PackageFiles {
+  /**
+   * Fetch package metadata from the npm registry.
+   *
+   * See https://github.com/npm/registry/blob/master/docs/responses/package-metadata.md
+   */
+  getPackageMetadata(packageName: string): Promise<Package> {
+    return npmFetch.json(
+      `/${packageName}`
+    ) as Promise<unknown> as Promise<Package>;
```
rictic, line 29: What's with the official types? Oh I see, they don't even see what API endpoint you're hitting, so they have no idea what type to return

packages/catalog-server/test.sh
```diff
@@ -0,0 +1 @@
+NODE_OPTIONS='--enable-source-maps' FIRESTORE_EMULATOR_HOST=localhost:8088 uvu test "_test\.js$"
```
rictic, line 1: what starts the firestore emulator?

rictic (commented):
A handful of comments / questions from my first pass

rictic (commented):
A handful of comments / questions from my first pass

packages/catalog-api/src/lib/schema.graphql
justinfagnani: Yeah. This is what I'm working on right now. We only know the dist-tags when we import a _package_, and then because we denormalize we have to update all _package versions_ and all custom elements that had a dist-tag change. Since we store the dist-tags, we can calculate if/when we need to do that.

packages/catalog-api/src/lib/schema.graphql
justinfagnani: I need to document it here. It's `{package-name}/{module-path}#{export-name}` where export name is a JS identifier for JS exports and a custom element name for custom element exports.

The code is here: https://github.com/webcomponents/webcomponents.org/blob/main/packages/custom-elements-manifest-tools/src/lib/reference-string.ts

packages/catalog-server/src/lib/firestore/custom-element-converter.ts
justinfagnani: "client side" meaning the Firestore client running on the server, yes.

packages/catalog-server/src/lib/npm.ts
justinfagnani: There are some packages with types for the two API endpoints, but I haven't been able to tell which is best... it does seem like this package should be typed and assume or validate that you're hitting a complaint API endpoint though.

packages/catalog-server/test.sh
justinfagnani: The `firebase emulators:exec` call here: https://github.com/webcomponents/webcomponents.org/pull/1323/files#diff-0caa62d7c8093d80e6beceaaea2ea9dc234094eeec43e2a036bfbc1dcbdeb08eR54

justinfagnani:
Thanks @rictic ! I'll update the comment, and I have another PR I'm about to push that improves the schema types like we were talking about.

packages/catalog-api/src/lib/schema.graphql
```diff
@@ -0,0 +1,246 @@
...
+  author: String!
+
+  tagName: String
+
+  className: String
+
+  """
+  Reference to the custom element export.
+
+  Reference strings have the format: `{package-name}/{module-path}#{tag-name}`
```
rictic, line 218: What do you think about making reference a type rather than sending down strings?

rictic (changes requested):
Let's configure CI to test with a real firestore instance. I'm confident that tests for important scenarios will get written, so my priority here is just that we have a test environment that's real enough to catch problems before production

The emulator is nice, but I don't trust like that

I don't trust like that

packages/catalog-api/src/lib/schema.graphql
```diff
@@ -0,0 +1,246 @@
...
+  """
+  Validates a package, returning a list of problems
+  """
+  validate(packageName: String!, tag: String = "latest"): [String!]!
+}
+
+type Mutation {
+  importPackage(packageName: String!): PackageInfo
+
+  deletePackage(packageName: String!): Boolean
```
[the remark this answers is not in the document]
justinfagnani: Yeah. I'll do that, but also delete this from the schema for now (it's not wired up and probably shouldn't ever exist in this form).

packages/catalog-api/src/lib/schema.graphql
justinfagnani: I'm open to it... though it makes GraphQL API calls more verbose. Like on the GraphQL level there's a question of whether you'd ever want to select one of the fields without the others. There's no default "give me all the fields" syntax. I also think that defining a string format will likely be useful elsewhere, like in URLs.

justinfagnani:
@rictic 
> Let's configure CI to test with a real firestore instance.

I'm ok with this too, but have to figure out how to do it. The nice thing about the emulator is that you start with a fresh database each time it starts. If we test on real Firestore we'll have to figure out how to reliably wipe the database. Also, I have not deployed any version of this code to GCP yet. I'll have to create a new test-only GCP project too.

justinfagnani:
Regarding real Firestore: there is a bit of work to do here, both on the testing side and on getting the GCP deployments setup for the first time.

I filed https://github.com/webcomponents/webcomponents.org/issues/1332 to track the testing setup and code changes required. I'd rather not block PRs that have emulated tests, which in theory should be accurate enough for unit testing - the Firestore docs say so - and work in parallel towards preview and test GCP deployments. @rictic

packages/catalog-api/src/lib/schema.graphql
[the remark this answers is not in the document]
justinfagnani: I removed this, but I'll add the comment to a query method that's not implemented yet. Keeping the method rather than removing specifically to get the comment started.

rictic:
> If we test on real Firestore we'll have to figure out how to reliably wipe the database.

I don't think that's a good idea, because it would mean multiple tests running at the same time would break each other. Better to lean into the namespaces work, but use random namespaces when in test.

rictic:
I'm ok punting the use of a real firestore as long as we have agreement that we should be running all tests in CI with the real firestore before launch.

if we don't have agreement we should talk more :)

justinfagnani:
> > If we test on real Firestore we'll have to figure out how to reliably wipe the database.
> 
> I don't think that's a good idea, because it would mean multiple tests running at the same time would break each other. Better to lean into the namespaces work, but use random namespaces when in test.

This is exactly what I mean. The clean up function would clean all collections tagged with the namespace used in _that_ test run. The test setup would generate a namespace used by the test and cleanup script. This would be done specifically to isolate test runs and enable parallelism.

> I'm ok punting the use of a real firestore as long as we have agreement that we should be running all tests in CI with the real firestore before launch.
> 
> if we don't have agreement we should talk more :)

Very much agreement. The namespacing, parallelism, and cleanup work can be developed against both local emulator or the GCP test project, and we can hit either from local dev or CI. As soon as we have the GCP infra ready used with the switch of a flag in a YAML file.

packages/catalog-api/package.json
```diff
@@ -0,0 +1,72 @@
+{
+  "name": "@webcomponents/catalog-api",
+  "version": "0.0.0",
+  "private": true,
+  "description": "Schema and interfaces for the @webcomponents/catalog-server",
+  "author": "Google LLC",
+  "homepage": "",
+  "license": "APACHE-2.0",
```
aomarks, line 8: Are we doing Apache or BSD?

packages/catalog-api/package.json
```diff
@@ -0,0 +1,72 @@
...
+        "out/"
+      ],
+      "output": [
+        "lib",
+        "index.{js,js.map,d.ts,d.ts.map}"
+      ],
+      "clean": "if-file-deleted"
+    },
+    "build:ts": {
+      "command": "../../node_modules/.bin/tsc",
```
aomarks, line 33: why do we keep doing this instead of just `tsc`? npm will resolve it fine, right?

packages/catalog-server/src/lib/catalog.ts
```diff
@@ -0,0 +1,134 @@
...
+export interface CatalogInit {
+  repository: Repository;
+  files: PackageFiles;
+}
+
+/**
+ * Implements operations for reading and writing to a catalog.
+ */
+export class Catalog {
+  private repository: Repository;
```
aomarks, line 23: We can use real #private fields

packages/catalog-server/src/lib/catalog.ts
```diff
@@ -0,0 +1,134 @@
...
+    );
+    console.log('done');
+    return {problems};
+  }
+
+  /**
+   * Returns the package version metadata, custom elements, and problems
+   * @param packageName
+   * @param version
+   * @returns
```
aomarks, line 107:
```suggestion

```

packages/catalog-server/src/lib/firestore/custom-element-converter.ts
```diff
@@ -0,0 +1,32 @@
...
+      distTags: snapshot.get('distTags'),
+      author: snapshot.get('author'),
+      tagName: snapshot.get('tagName'),
+      className: snapshot.get('className'),
+      customElementExport: snapshot.get('customElementExport'),
+      declaration: snapshot.get('declaration'),
+    };
+  },
+  toFirestore(_packageInfo: CustomElement) {
+    throw new Error('not implemented');
```
aomarks, line 30: Why doesn't this need to be implemented?

packages/catalog-server/src/lib/firestore/firestore-repository.ts
```diff
@@ -0,0 +1,323 @@
...
+      const distTags = packageMetadata['dist-tags'];
+      const versionDistTags = getDistTagsForVersion(distTags, version);
+
+      // Store package data and mark version as ready
+      // TODO: we want this type to match the schema and converter type. How do
+      // we get strongly typed refs?
+      await t.set(versionRef, {
+        status: VersionStatus.READY,
+        lastUpdate: FieldValue.serverTimestamp(),
+        // package: packageName,
```
aomarks, line 92: Is this supposed to be commented? package seems important

packages/catalog-server/tsconfig.json
```diff
@@ -0,0 +1,30 @@
...
+    "experimentalDecorators": true,
+    "skipLibCheck": true,
+    "forceConsistentCasingInFileNames": true 
+  },
+  "include": ["src/**/*.ts"],
+  "exclude": [],
+  "references": [
+    { "path": "../catalog-api" },
+    { "path": "../custom-elements-manifest-tools" }
+  ]
```
aomarks, line 29: This precludes the use of wireit. Intentional?

aomarks (changes requested):

packages/catalog-server/src/lib/firestore/custom-element-converter.ts
justinfagnani: I'm in the middle of more consistently using converters for both read and write operations. I currently don't use them for writing custom elements, but wanted to throw here in case it was used that way so we wouldn't just drop all the data.

packages/catalog-server/src/lib/firestore/firestore-repository.ts
justinfagnani: It's not in the GraphQL schema type for PackageVersion, since PackageVersion is always nested under a PackageInfo in the GraphQL schema and is a subcollection of package in the Firestore layout. So the package name is always available there.

What's it's needed for is denormalization for queries, which aren't implemented yet. I can remove or add a comment that this needs to be added and a type created with denormalized fields added to the schema type.

packages/catalog-server/tsconfig.json
justinfagnani: I'm trying to preserve jump-to-definition across packages. Builds via Wireit seem to work fine with this.

packages/catalog-api/package.json
justinfagnani: Oh, if that works, then yeah, I'll change it.

packages/catalog-api/package.json
justinfagnani: I'm not entirely sure... let's chat?

packages/catalog-server/src/lib/catalog.ts
justinfagnani: Ok, as long as we have a target of es2022 or newer. I don't want to compile to WeakMaps.

packages/catalog-server/tsconfig.json
aomarks: I guess it will be ok as of recent Wireit versions, actually. Here's the situation I was worried about:

- Project A is in state 1
- Build project A with wireit
- Modify project A to state 2
- Build project B with wireit, where B references A, causing project A to be built without wireit knowing
- Modify project A back to state 1
- Build project A with wireit

This is potentially a bad state: wireit thinks the output is already correct, since as far as it knows project A was already built at state 1. But actually it's built for state 2. But, since we now do output verification, this scenario will actually work out fine. Wireit will say "hey! the output changed" and do a fresh build.

aomarks: Err, there could be some weird similar situations, or races that happen where wireit thinks it has a lock on building project A, but actually project B is simultaneously implicitly building it without wireit knowing...

I guess my guess it is this will be fine the vast majority of the time, but there is a small chance of something getting weird sometime.

justinfagnani: But don't you specifically have to pass `--build` to `tsc` to build referenced projects? w/o that you just get incremental and referenced project support in IDEs.

packages/catalog-server/src/lib/firestore/firestore-repository.ts
justinfagnani: Added comment

packages/catalog-server/src/lib/catalog.ts
justinfagnani: Set target to es2022 and changed to real private fields.

packages/catalog-api/package.json
justinfagnani: The old site is using Apache: https://github.com/webcomponents/webcomponents.org/blob/old-site/LICENSE and I copied the license over to the main branch in the initial commit.

But this is all new code, so I'm not sure if we need to stick with that. It seems easier though.

aomarks: Since Lit will be in our bundles, it would be very slightly preferable for compression to have the same license in the client code that we ship. And I think it simplifies this entire repo if the entire thing has one license, because otherwise we need multiple LICENSE files, etc.

packages/catalog-server/tsconfig.json
aomarks: Oh, if that's the case then that's great. Are we sure about it? I haven't tested.

justinfagnani: https://www.typescriptlang.org/docs/handbook/project-references.html#caveats-for-project-references

> Additionally, to preserve compatibility with existing build workflows, tsc will not automatically build dependencies unless invoked with the --build switch.

Seems ok. I thought the only problem was you measuring `--build` to be considerably faster than just incremental builds, which seems like either a bug or just really unintuitive. I'm personally willing to tradeoff a slightly slower build to have working IDE features, but maybe esbuild means we don't have to make that tradeoff even.

packages/catalog-api/src/lib/schema.graphql
rictic: If it's a string we'll want to provide an API for clients to use to properly construct a reference (mainly to handle escaping)

rictic (approved):

packages/catalog-api/package.json
aomarks, line 8:
```suggestion
  "license": "Apache-2.0",
```

packages/catalog-api/src/lib/schema.graphql
```diff
@@ -0,0 +1,250 @@
+type Query {
```
aomarks, line 1: I think we still have an exercise to talk through what the best API should be exactly, but I'm fine with landing this knowing that we can continue to iterate on it.

packages/catalog-server/firestore.rules
```diff
@@ -0,0 +1,16 @@
...
+      // This rule allows anyone with your database reference to view, edit,
+      // and delete all data in your database. It is useful for getting
+      // started, but it is configured to expire after 30 days because it
+      // leaves your app open to attackers. At that time, all client
+      // requests to your database will be denied.
+      //
+      // Make sure to write security rules for your app before that time, or
+      // else all client requests to your database will be denied until you
+      // update your rules.
+      allow read, write: if request.time < timestamp.date(2022, 7, 23);
```
aomarks, line 13: Err... ok for now I guess?

packages/catalog-server/package.json
```diff
@@ -0,0 +1,80 @@
+{
+  "name": "@webcomponents/catalog-server",
+  "version": "0.0.0",
+  "private": true,
+  "description": "A web component catalog database and server",
+  "author": "Google LLC",
+  "homepage": "",
+  "license": "APACHE-2.0",
```
aomarks, line 8:
```suggestion
  "license": "Apache-2.0",
```

packages/catalog-server/tsconfig.json
aomarks: SGTM

aomarks (approved):

PR #1328, opened 2022-09-15
Schema types

justinfagnani:
Splits PackageInfo and PackageVersion types into Readable and Unreadable versions, depending on their import status. This lets us correctly model the fact that a successfully imported object has all of its required fields, and a failed import will only have status and lastUpdated fields.


packages/catalog-server/src/lib/firestore/firestore-repository.ts
```diff
@@ -234,7 +238,7 @@ export class FirestoreRepository implements Repository {
   async getPackageVersion(
     packageName: string,
     version: string
-  ): Promise<PackageVersion | undefined> {
+  ): Promise<Omit<PackageVersion, 'customElements' | 'problems'> | undefined> {
```
rictic, line 241: Did the type system catch that you needed to add this Omit, or did you manually have to notice that?

packages/catalog-server/src/test/lib/catalog_test.ts
```diff
@@ -49,12 +53,16 @@ test('Gets package version data from imported package', async () => {
   );
 
   assert.ok(result);
+  assert.equal(result.status, VersionStatus.READY);
+
+  const packageVersion = result as ReadablePackageVersion;
```
rictic, line 58: If this was written:

```typescript
if (result.status !== VersionStatus.READY) {
  throw new Error('...');
}
```

would this `as` assertion be unnecessary?

packages/catalog-api/src/lib/schema.graphql
```diff
@@ -8,12 +8,17 @@ type Query {
   Queries custom elements in the entire catalog, from the latest version of each
   package.
 
-  Eventually this will use some sort of ranking algorithm, otherwise the order
-  will just be defined by the database implementation.
+  Eventually this will have more query parameters, and use some sort of ranking
+  algorithm, otherwise the order will just be defined by the database
+  implementation.
   """
   elements(tag: String = "latest", limit: Int): [CustomElement!]!
```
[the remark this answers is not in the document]
justinfagnani: done

packages/catalog-api/src/lib/schema.graphql
```diff
@@ -28,13 +33,17 @@ type Mutation {
...
-or related to the latest version of the package.
+Information about an npm package that's independent of a specific version.
 """
-type PackageInfo {
+union PackageInfo = ReadablePackageInfo | UnreadablePackageInfo
+
+"""
+A package that has been successfuly imported
+"""
+type ReadablePackageInfo {
```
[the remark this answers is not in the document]
justinfagnani: This should not be possible. Readable means that the other fields have been filled out at least once and can be used to display data in a UI, even if they are stale. Because pakage versions are immutable, we can at least read something. I think even redacted packages will be a "readable" state, in that the data is there.

But given proper use of the Firestore document converters, if that case did happen (readable->unreadable) we would not return the additional fields of the readable type. This could be an area where we assert that database objects do not have extra fields.

packages/catalog-server/src/lib/firestore/firestore-repository.ts
justinfagnani: I did that because these fields map to sub collections and so aren't part of the PackageVersion document. There's more work to do here to make the mapping from GraphQL types to Firestore types more regularized. I had a utility type to do some of the mapping, but it required some manual configuration like listing the collections. I tried to see if graphql-codegen could adjust types based on field directives (like `@collection` and `@map` that I have in the schema) but it can't do that (yet).

So I'm thinking we should be more formalized here. We don't have _that_ many types in play and we can make a module with utility type like that, that also takes care of specifying the additional denormalized fields, and applies it to all the GraphQL schema types. Then we'll have names for all of these and code/documentation on the specific mapping between types.

packages/catalog-server/src/test/lib/catalog_test.ts
justinfagnani: Yeah... I just wish we could get uvu asserts to also be type guards.

packages/catalog-api/src/lib/schema.graphql
```diff
@@ -163,19 +168,38 @@ type PackageVersion {
...
+  The package version is indexed and read to be read.
   """
-  INITIALIZING
+  READY
 
   """
-  The version is indexed and read to be read.
+  A non-recoverable error. The package was imported, but invalid in some way
+  that prevents it from being displayed in the catalog. Since package versions
+  are immutable, this version cannot be fixed. A new package version will may
```
rictic, line 188: Typo: will may

rictic (approved):

aomarks (approved):

PR #1329, opened 2022-09-17
Add Catalog.importPackage()

justinfagnani:


packages/catalog-server/src/lib/firestore/firestore-repository.ts
```diff
@@ -255,73 +367,71 @@ export class FirestoreRepository implements Repository {
...
-          ? SchemaType[K] extends ReadonlyArray<infer T>
-            ? Omit<T, MapFields[K]>
-            : never
-          : MapFields[K];
-      }
-    : K extends Collections
-    ? never
-    : SchemaType[K];
-};
+// /**
```
rictic, line 399: Why comment this out?

justinfagnani: Oh. Not using it anymore, even though I do want to bring this back in some form. This is a type-only utility and I think I want to make an actual transformer function. I also was searching for a way to automate this from the GraphQL schema w/ field directives, but I think we'd have to write a plugin for that.

I'll delete it for now, but I'm likely to want to make a firestore/schema.ts module with something like this in it.

packages/catalog-server/src/lib/catalog.ts
```diff
@@ -31,8 +53,111 @@ export class Catalog {
...
+    try {
+      newPackage = await this.#files.getPackageMetadata(packageName);
+    } catch (e) {
+      await this.#repository.endPackageImportWithError(packageName);
+      return;
+    }
+
+    if (newPackage === undefined) {
+      await this.#repository.endPackageImportWithNotFound(packageName);
+      // TODO (justinfagnani): a crazy edge case would be a package that was
```
rictic, line 103: This isn't always a crazy error, some systems handle some internal errors by giving a 404 or similar rather than a 500 error

Hopefully the npm registry isn't one, but something to be mindful of

rictic (approved):

packages/catalog-server/src/lib/catalog.ts
```diff
@@ -31,8 +53,111 @@ export class Catalog {
...
+      return;
+    }
+
+    const newDistTags = newPackage['dist-tags'];
+
+    // This will be undefined if we don't need to import any new package
+    // version (if the 'latest' dist-tag didn't change).
+    let versionToImport: string | undefined = newDistTags['latest'];
+
+    if (isReadablePackage(currentPackageInfo)) {
```
aomarks, line 114: Now in context I'm finding this "readable" vs "unreadable" naming slightly unintuitive. What if we called it "isImported" and "isNotImported" or something?

packages/catalog-server/src/lib/firestore/firestore-repository.ts
```diff
@@ -4,13 +4,7 @@
  * SPDX-License-Identifier: BSD-3-Clause
```
aomarks, line 4: BSD or Apache? Same for all files, I just noticed the headers are not consistent with the package.json

packages/catalog-server/src/lib/firestore/package-info-converter.ts
```diff
@@ -9,33 +9,67 @@ import {
...
+ * packages.
+ */
+export const packageNameToId = (packageName: string) =>
+  packageName.replace('/', '__');
+
+/**
+ * Converts a Firestore ID to a package name.
+ */
+export const idToPackageName = (packageName: string) =>
+  packageName.replace('/', '__');
```
aomarks, line 75: Shouldn't this be the other way around?

(And shouldn't a test cover this somewhere?)

packages/catalog-server/src/lib/firestore/package-info-converter.ts
```diff
@@ -9,33 +9,67 @@ import {
...
+/**
+ * Converts a package name to a Firestore ID.
+ *
+ * Firestore IDs cannot include a '/', so we convert it to '__'.
+ *
+ * This is a similar to the transform TypeScript does for `@types`
+ * packages.
+ */
+export const packageNameToId = (packageName: string) =>
+  packageName.replace('/', '__');
```
aomarks, line 69: Just to confirm, `replace` is fine (vs `replaceAll`) because there can only be one "/" in a package name?

packages/catalog-server/src/lib/repository.ts
```diff
@@ -15,10 +15,46 @@ import type {Package} from '@webcomponents/custom-elements-manifest-tools/lib/np
...
+ * implementations of Repository so the catalog can run on different local and
+ * cloud environments.
+ *
+ * Repository methods should be as simple as possible. They should either
+ * operate on a single entity-kind/table (and let Catalog coordinate access to
+ * multiple entity-kinds) or use a transaction to guarantee consistency across
+ * multiple entity-kinds.
+ *
+ * Some workflows, like importing packages and versions, are split across
+ * several methods and do not allow for a single transaction. These worflows do
```
aomarks, line 31:
```suggestion
 * several methods and do not allow for a single transaction. These workflows do
```

aomarks (approved):

packages/catalog-api/codegen.yml
```diff
@@ -13,6 +13,7 @@ generates:
       namingConvention: keep
       noSchemaStitching: true
       operationResultSuffix: Data
+      # TODO: try false
```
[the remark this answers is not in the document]
justinfagnani: If you look at the generated TS, there are lots of utility types and I'm wondering how much more simple we can get them to be.

packages/catalog-server/src/lib/catalog.ts
justinfagnani: That is true. I'll look into it

packages/catalog-server/src/lib/catalog.ts
justinfagnani: I'm open. To me the meant meant "can I read useful data from it", but "unreadable" sounds a bit like an I/O error.

packages/catalog-server/src/lib/firestore/package-info-converter.ts
justinfagnani: Exactly. This is also the same transform that TypeScript uses in the `@types` packages for scoped sub-packages.

packages/catalog-server/src/lib/firestore/package-info-converter.ts
justinfagnani: yes, and yes. We're not using the returned package name yet. I added an assertion in catalog_test, but we need more test packages that include scoped names. In the next PR I'm importing Shoelace's manifest to test. We should do that for a few other real-world packages.

packages/catalog-server/src/lib/firestore/firestore-repository.ts
justinfagnani: Everything should be Apache now

PR #1334, opened 2022-09-24
Set up separate dev and prod builds with esbuild

aomarks:
**Note to reviewers**: I first used tsc + Rollup in this PR, and then replaced it with esbuild, so you'll see those commits in the middle. You can ignore them, or read them if you're interested.

Sets up separate dev and prod eleventy builds for the new webcomponents.org. Both builds use esbuild.

Running `npm run watch:dev` or `npm run watch:prod` from the top-level will give you live refresh versions of either build.

The dev build is super fast because [1] it uses symlinks so that Eleventy doesn't need to rebuild when static assets or JS changes, and [2] it runs esbuild with no minification or bundling.

There is a separate `check:types` command in the `client` package which runs TypeScript in `noEmit` mode. This runs in parallel when you use either of the top-level `watch:*` commands. So the experience is that the browser updates immediately, and a few moments later you also get any type errors.

This design lets us use identical `<script>` tags in both dev and prod mode, which is nice. On lit.dev we needed to generate different script tags for each mode.

Also addresses the post-merge comments from https://github.com/webcomponents/webcomponents.org/pull/1333


packages/client/package.json
```diff
@@ -0,0 +1,56 @@
...
+  "wireit": {
+    "build": {
+      "dependencies": [
+        "build:dev",
+        "build:prod",
+        "check:types"
+      ]
+    },
+    "build:dev": {
+      "command": "esbuild --format=esm --target=es2022 --color --outdir=lib src/**/*.ts ",
```
justinfagnani, line 24: Are we using esbuild here just because it's faster than `tsc`?

How quickly does it keep up with TypeScript? When could we start using 4.9 and the `satisfies` operator?

packages/client/package.json
```diff
@@ -0,0 +1,56 @@
...
+  },
+  "wireit": {
+    "build": {
+      "dependencies": [
+        "build:dev",
+        "build:prod",
+        "check:types"
+      ]
+    },
+    "build:dev": {
```
justinfagnani, line 23: Could we have a command that does `build:dev` and `check:types`, but not `build:prod`?

packages/content/eleventy.config.dev.cjs
```diff
@@ -0,0 +1,31 @@
...
+  symlinkForce('../../client/lib/entrypoints', '_dev/js');
+  return {
+    dir: {
+      input: 'site',
+      output: '_dev',
+    },
+  };
+};
+
+function symlinkForce(target, path) {
```
justinfagnani, line 21: Add comment on why this is here?

packages/content/package.json
```diff
@@ -7,21 +7,36 @@
...
       "output": [
         "_site"
       ],
       "dependencies": [
-        "../components:build"
+        "../client:build:prod"
+      ]
+    },
+    "build:dev": {
+      "command": "eleventy --config=eleventy.config.dev.cjs",
```
justinfagnani, line 28: should this have a dependency on `../client:build:dev`?

packages/client/package.json
aomarks: Yeah, esbuild is just SO much faster than tsc. It's so much better for the dev cycle, because you save your ts file, and the browser instantly updates. (Right now we don't have a lot of code, so the difference is small, but over time the difference will become large).

I don't know how quickly it keeps up with TypeScript. I would guess pretty good, but yeah that's something to keep in mind. An option we do have down the line is to use esbuild for dev code, and TypeScript for the prod build, e.g. if we preferred the output for some reason (but that doesn't address unsupported syntax).

packages/client/package.json
aomarks: What would you call it?

packages/content/package.json
aomarks: No, because otherwise every time the client code changes, Eleventy will re-run unnecessarily, because in dev mode that's just a symlink.

packages/content/eleventy.config.dev.cjs
aomarks: Done

packages/client/package.json
justinfagnani: I'd be tempted to call it `build:dev` and this one `build:ts:dev` with tests depending on `build:dev`.

packages/content/eleventy.config.dev.cjs
justinfagnani: Did you upload the commit?

packages/content/package.json
justinfagnani: Hmm... seems like that leaves a situation where the client output could be stale. Is there a different type of dependency at play here. Like: when you run task A, ensure B is up to date, but when you run B, do _not_ dirty A.

packages/content/eleventy.config.dev.cjs
justinfagnani: I see the comment in #1335... that's fine.

packages/client/package.json
justinfagnani: I see this also in #1335 ✔️

justinfagnani (approved):

PR #1335, opened 2022-09-25
Set up lit analyzer, eslint, and prettier checks

aomarks:
Integrates lit analyzer, eslint, and prettier checks.

Running `check` runs them all, and they also run during `watch:dev` and `watch:prod`.


.eslintrc.json
```diff
@@ -0,0 +1,23 @@
...
+    "ecmaVersion": 2022,
+    "sourceType": "module"
+  },
+  "plugins": [
+    "@typescript-eslint",
+    "eslint-plugin-no-only-tests",
+    "eslint-plugin-import"
+  ],
+  "rules": {
+    "import/extensions": ["error", "always"],
```
justinfagnani, line 18: Should we enable `import/no-extraneous-dependencies` too? Some packages like custom-elements-manifest-tools may be published to npm, so it'll matter.

package.json
```diff
@@ -12,17 +12,26 @@
     "build:prod": "wireit",
     "serve:dev": "wireit",
     "serve:prod": "wireit",
-    "watch:dev": "npm run build:dev && (npm run serve:dev & npm run build:dev --watch & npm run check --watch)",
-    "watch:prod": "npm run build:prod && (npm run serve:prod & npm run build:prod --watch & npm run check --watch)",
+    "watch:dev": "npm run build:dev && (npm run serve:dev > /dev/null 2>&1 & WIREIT_FAILURES=continue npm run build-and-check:dev --watch)",
+    "watch:prod": "npm run build:prod && (npm run serve:prod > /dev/null 2>&1 & WIREIT_FAILURES=continue npm run build-and-check:prod --watch)",
```
justinfagnani, line 16: Do we really need watch:prod? When do we think we'd use it?

package.json
```diff
@@ -58,10 +67,40 @@
...
+        "check:format"
       ]
+    },
+    "check:lint": {
+      "command": "eslint --color --cache --cache-location .eslintcache \"packages/client/src/**/*.ts\"",
+      "files": [
+        ".eslintrc.json",
+        "packages/client/src/**/*.ts"
+      ],
+      "output": []
```
justinfagnani, line 95: Should `.eslintcache` be included in output?

package.json
```diff
@@ -58,10 +67,40 @@
...
+    },
+    "check:format": {
+      "command": "prettier \"./packages/client/src/**/*.ts\" --check --cache --cache-strategy=metadata",
+      "files": [
+        ".prettierrc.json",
+        "packages/client/src/**/*.ts"
+      ],
+      "output": []
```
justinfagnani, line 103: Include cache file here too?

justinfagnani (approved):

package.json
aomarks: I don't see why not. I think there will end being potentially significant differences that having a fast debug cycle will be useful for, e.g. `prod` will likely use Eleventy SSR, but `dev` won't.

package.json
aomarks: See above.

package.json
aomarks: Nope, that would cause it to be deleted every time, making it useless. Unlike with `.tsbuildinfo`, we don't have the problem where deleting an input file leaves the artifact around, because this doesn't remove anything. That's the reason we put `.tsbuildinfo` in the `output` usually.

.eslintrc.json
aomarks: Done

Issue #1326, opened 2022-09-14
[site-server] Add new site-server package with a simple server


justinfagnani:
The webcomponents.org site will have a server separate from the catalog's GraphQL server. This server will serve the content generated in a `site-content` package with 11ty.

PR #1342, opened 2022-09-28
[site-server] Add a production server

aomarks:
Just a minimal static Koa server

Also fixes the SPDX identifiers in client .ts files to be Apache-2.0 instead of BSD-3-Clause

Fixes https://github.com/webcomponents/webcomponents.org/issues/1326


packages/server/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@webcomponents/internal-site-server",
```
justinfagnani, line 2: I would probably name the folder `site-server` to be more clear when browsing the FS and also seeing `catalog-server`

packages/server/src/prod-server.ts
```diff
@@ -0,0 +1,51 @@
...
+import Koa from 'koa';
+import koaStatic from 'koa-static';
+import koaConditionalGet from 'koa-conditional-get';
+import koaEtag from 'koa-etag';
+import {fileURLToPath} from 'url';
+import * as pathlib from 'path';
+
+const PORT = process.env['PORT'] || 8080;
+const STATIC_ROOT = pathlib.resolve(
+  fileURLToPath(import.meta.url),
```
justinfagnani, line 16: super nit: might be nice to use `fileURLToPath(new URL('../../../context/_site', import.meta.url))`

justinfagnani (approved):

packages/server/package.json
aomarks: Done

packages/server/src/prod-server.ts
aomarks: Done (removes need for path library import :-1: )

PR #1343, opened 2022-09-30
Add GraphQL endpoint

### LINKED ISSUE #1325: [catalog-server] Add initial GraphQL endpoint, resolvers and GraphiQL
The most basic usable GraphQL starting point is a `/graphql` route that loads the `GraphiQL` interface for GET requests and performs queries for POST requests.

justinfagnani:
This adds a Koa server and `/graphql` endpoint to the `catalog-server` package.

The server is implemented with [graphql-helix](https://github.com/contra/graphql-helix), which makes it very easy to parse a GraphQL request, execute it against a schema, and send the result. I chose this after koa-graphql [seemed unmaintained](https://github.com/graphql-community/koa-graphql/issues/142) and have been happy with it.

GraphQL execution is implemented with "resolvers" - objects that let the GraphQL system ask for objects and fields. There are resolvers based on type like `PackageInfo`, `ReadablePackageInfo`, `PackageVersion`, etc. Then field resolvers for sub-collections like `ReadablePackageVersion.customElements`. The resolvers use the Catalog interface to fetch results. Resolvers are combined with the parsed schema into a "executable schema" which graphql-helix uses to validate and perform the queries.

There is no access control, rate-limiting, etc., yet.

Fixes #1325


packages/catalog-server/index.js
```diff
@@ -0,0 +1,12 @@
+/**
```
aomarks, line 1: Should not be checked in

packages/catalog-server/index.js.map
```diff
@@ -0,0 +1 @@
+{"version":3,"file":"index.js","sourceRoot":"","sources":["src/index.ts"],"names":[],"mappings":"AAAA;;;;GAIG;AAGH,OAAO,EAAC,UAAU,EAAC,MAAM,iBAAiB,CAAC;AAE3C,MAAM,GAAG,GAAG,MAAM,UAAU,EAAE,CAAC;AAC/B,MAAM,MAAM,GAAG,GAAG,CAAC,MAAM,CAAC,IAAI,EAAE,GAAG,EAAE;IACnC,MAAM,IAAI,GAAI,MAAM,CAAC,OAAO,EAAkB,CAAC,IAAI,CAAC;IACpD,OAAO,CAAC,GAAG,CAAC,oCAAoC,IAAI,EAAE,CAAC,CAAC;AAC1D,CAAC,CAAC,CAAC"}
```
aomarks, line 1: Should not be checked in

packages/catalog-server/src/lib/catalog.ts
```diff
@@ -84,25 +93,27 @@ export class Catalog {
...
+    console.log('Fetching package metadata...');
     let newPackage: Package | undefined;
     try {
       newPackage = await this.#files.getPackageMetadata(packageName);
     } catch (e) {
       await this.#repository.endPackageImportWithError(packageName);
-      return;
+      return {};
     }
+    console.log(' done');
```
aomarks, line 110: With concurrency, these "done" logs might be hard to correlate to the action they are associated with.

packages/catalog-server/src/lib/catalog.ts
```diff
@@ -144,16 +155,43 @@ export class Catalog {
...
     if (versionToImport !== undefined) {
-      await this.importPackageVersion(packageName, versionToImport);
+      importResult = await this.importPackageVersion(
+        packageName,
+        versionToImport
+      );
     }
+
+    console.log('Marking package ready...');
+    const newPackageInfo: ReadablePackageInfo = {
```
aomarks, line 179: Just to clarify -- these changes weren't in the PR description -- we weren't setting the ready state before?

packages/catalog-server/src/lib/firestore/firestore-repository.ts
```diff
@@ -219,23 +220,28 @@ export class FirestoreRepository implements Repository {
       const versionDistTags = getDistTagsForVersion(distTags, version);
 
       // Store package data and mark version as ready
-      await t.set(versionRef, {
+      // TODO: make converter handle denormalized data
```
aomarks, line 223: what is denormalized data in this context?

packages/catalog-server/src/lib/firestore/firestore-repository.ts
```diff
@@ -364,13 +370,34 @@ export class FirestoreRepository implements Repository {
...
+
+    if (versionNumber !== null) {
+      // If version is valid semver we can build a document reference.
+      const versionRef = this.getPackageVersionRef(packageName, versionNumber);
+      const versionDoc = await versionRef.get();
+      return versionDoc.data();
+    } else {
+      // If version is not a valid semver it may be a dist-tag
+
+      // First, filter out semver ranges, since npm doesn't allow semver
```
aomarks, line 385: Are we not going to support semver ranges in our API?

packages/catalog-server/src/lib/firestore/package-info-converter.ts
```diff
@@ -47,10 +58,11 @@ export const packageInfoConverter: FirestoreDataConverter<
         // FieldValue.serverTimestamp() here.
         lastUpdate: data.lastUpdate,
         description: data.description,
-        distTags: new Map(
-          // We don't support FieldValues in distTags, so cast away:
-          (data.distTags as DistTag[]).map((t) => [t.tag, t.version])
-        ),
+        distTags: distTagListToMap(data.distTags as Array<DistTag>),
+        // new Map(
```
aomarks, line 62: Remove commented code?

packages/catalog-server/src/lib/firestore/package-info-converter.ts
```diff
@@ -17,26 +17,37 @@ import {
...
-      distTags: graphQLDistTags,
-      // `version` is left to a sub-collection query
-    } as ReadablePackageInfo;
+export const packageInfoConverter: FirestoreDataConverter<PackageInfo> = {
+  fromFirestore(snapshot: QueryDocumentSnapshot<DocumentData>): PackageInfo {
+    const name = idToPackageName(snapshot.id);
+    const status = snapshot.get('status');
+    const lastUpdate = (snapshot.get('lastUpdate') as Timestamp).toDate();
+
+    if (status === ReadablePackageStatus.READY || status === ReadablePackageStatus.UPDATING) {
```
aomarks, line 31: Use the `isReadablePackage` utility?

packages/catalog-server/src/lib/graphql.ts
```diff
@@ -0,0 +1,127 @@
...
+ * Creates a GraphQLSchema object with resolvers
+ */
+export const makeExecutableCatalogSchema = async (catalog: Catalog) => {
+  const schemaPath = require.resolve(
+    '@webcomponents/catalog-api/src/lib/schema.graphql'
+  );
+  const schemaSource = await readFile(schemaPath, 'utf8');
+
+  // Resolvers for our schema. See https://graphql.org/learn/execution/ for
+  // an explaination of the role of resolvers in performing GraphQL queries.
```
aomarks, line 25:
```suggestion
  // an explanation of the role of resolvers in performing GraphQL queries.
```

packages/catalog-server/src/lib/graphql.ts
```diff
@@ -0,0 +1,127 @@
...
+        } else {
+          console.log(`package ${packageName} found in db and not readable`);
+          return null;
+        }
+      },
+    },
+    PackageInfo: {
+      __resolveType(obj) {
+        if (isReadablePackage(obj)) {
+          return 'ReadablePackageInfo';
```
aomarks, line 60: Are these strings type checked?

packages/catalog-server/src/lib/graphql.ts
```diff
@@ -0,0 +1,127 @@
...
+export const makeExecutableCatalogSchema = async (catalog: Catalog) => {
+  const schemaPath = require.resolve(
+    '@webcomponents/catalog-api/src/lib/schema.graphql'
+  );
+  const schemaSource = await readFile(schemaPath, 'utf8');
+
+  // Resolvers for our schema. See https://graphql.org/learn/execution/ for
+  // an explaination of the role of resolvers in performing GraphQL queries.
+  const resolvers: Resolvers = {
+    Query: {
```
aomarks, line 27: So `Query` is a special thing meaning the entrypoint query. Do we only ever have 1 entrypoint query? It seems like this resolver is assuming that all queries include a packageName, but we'll also support lookups by tag name, right?

packages/catalog-server/src/lib/graphql.ts
```diff
@@ -0,0 +1,127 @@
...
+      },
+    },
+    ReadablePackageVersion: {
+      customElements: async (
+        packageVersion,
+        {tagName}: {tagName?: string | null},
+        context,
+        _info
+      ) => {
+        // const packageName = packageVersion.name;
```
aomarks, line 101: Remove commented code

packages/catalog-server/src/lib/graphql.ts
```diff
@@ -0,0 +1,127 @@
...
+        return catalog.getCustomElements(
+          packageName,
+          version,
+          tagName ?? undefined
+        );
+      },
+    },
+    Mutation: {
+      async importPackage(_parent, {packageName}: {packageName: string}) {
+        console.log('mutation deletePackage', packageName);
```
aomarks, line 114: Not sure I understand this -- why do we log deletePackage and then just return a package?

aomarks (changes requested):
No tests :(

aomarks (approved):
Nice!

LGTM assuming comments addressed where relevant. We definitely need tests... but a followup PR is fine.

packages/catalog-server/src/lib/graphql.ts
justinfagnani: Yeah, `__resolveType` has to return a `typeof __typename` for the union `PackageInfo`.

packages/catalog-server/src/lib/graphql.ts
justinfagnani: Query and Mutation are special "root" types. Right now we're only implementing the `package` "field" (which is also a query). We will add more queries to the schema and resolves as we need them. The `elements()` query that's in the schema (and marked as not implemented) will be one of the main queries. This query is for accessing specific packages/versions/custom-elements in a top-down manner. `elements()` will be for actually searching across the whole catalog.

packages/catalog-server/src/lib/graphql.ts
justinfagnani: Add a TODO comment

packages/catalog-server/src/lib/graphql.ts
justinfagnani: Oh, this is leftover code from the prototype. I'll remove this.

justinfagnani: Actually, I think this is a good point to put some explicit import extras, like reporting problems. I'll add a TODO and comment in here.

packages/catalog-server/src/lib/catalog.ts
justinfagnani: Very true. We need to decide how much we want to log and how to do so. I've never use GCP Logging, but it looks like it has nice analytics features that could come in handy: https://cloud.google.com/logging

packages/catalog-server/src/lib/catalog.ts
justinfagnani: Yeah... need to add tests!

packages/catalog-server/src/lib/firestore/firestore-repository.ts
justinfagnani: The fields copied from PackageInfo like the packageName and distTags... This is actually already handled by those fields being in the GraphQL schema. Not sure fi we need to do more, so I'll remove the comment.

packages/catalog-server/src/lib/firestore/firestore-repository.ts
justinfagnani: I wasn't planning on it, I'm not sure what the use case is. And if we did, we'd need to figure out how to query (starting with a lexicographic range and filtering?) and in which gql queries it would make sense to enable.

packages/catalog-server/src/lib/firestore/package-info-converter.ts
justinfagnani: This is not a PackageInfo object, it's a Firestore DocumentSnapshot.

PR #1345, opened 2022-09-30
Add get-custom-elements_test and fix resolveReference

justinfagnani:
Fixes an issue with the Shoelace manifest where local references (to inside the same module) included a module field so we treated them as external references and failed to resolve non-exported declarations. Now we check if the module we're importing from and the reference's module are the same, and if so treat it as a local reference.

### DIFF
--- a/packages/custom-elements-manifest-tools/src/test/lib/get-custom-elements_test.ts
+++ b/packages/custom-elements-manifest-tools/src/test/lib/get-custom-elements_test.ts
@@ -1,0 +1,96 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {readFile} from 'fs/promises';
+import {fileURLToPath} from 'url';
+
+import {suite} from 'uvu';
+import * as assert from 'uvu/assert';
+
+import type {Package} from 'custom-elements-manifest/schema.js';
+import {getCustomElements} from '../../lib/get-custom-elements.js';
+
+const test = suite('get-custom-elements');
+
+const manifest: Package = {
+  schemaVersion: '1.0.0',
+  readme: 'README.md',
+  modules: [
+    {
+      kind: 'javascript-module',
+      path: 'my-project/my-element.js',
+      declarations: [
+        {
+          kind: 'class',
+          customElement: true,
+          tagName: 'my-element',
+          description: 'This is the description of the class',
+          name: 'MyElement',
+          members: [],
+        },
+      ],
+      exports: [
+        {
+          kind: 'js',
+          name: 'MyElement',
+          declaration: {
+            name: 'MyElement',
+          },
+        },
+        {
+          kind: 'custom-element-definition',
+          name: 'my-element',
+          declaration: {
+            name: 'MyElement',
+          },
+        },
+      ],
+    },
+  ],
+};
+
+test('Gets an element from a valid manifest', () => {
+  const customElements = getCustomElements(manifest, 'test', '0.0.0');
+  assert.equal(customElements.length, 1);
+  assert.equal(customElements[0]?.declaration.name, 'MyElement');
+  assert.equal(customElements[0]?.export.name, 'my-element');
+  assert.equal(customElements[0]?.module.path, 'my-project/my-element.js');
+  assert.equal(customElements[0]?.package, manifest);
+});
+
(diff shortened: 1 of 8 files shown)


justinfagnani:
I can't figure out why the shoelace manifest is not being found on CI 🤔

justinfagnani:
@aomarks do you have any ideas what could be going on here?

The `fs` functions aren't finding the `test/test-data/shoelace-2.0.0-beta.83.json` file that is included in the commits. The `test-data` directory just appears empty, but listing it does not raise an error, so it exists. I'm very confused.

justinfagnani:
I tried the test locally on a fresh checkout, and it fails there too. The wild part is that the test somehow deletes the file. I though it might be a wireit config, but if I skip the test the file isn't deleted, so I think that rules out wireit which has no knowledge of the implementation inside the test module.

Continuing with the weirdness, the file is gone by the time the debug code I added that lists the `test-data` directory runs. Nothing by then even mentioned that file by name. The only things are `fileURLToPath(import.meta.url)`, `process.cwd()`, `path.resolve()` and `fs.readdir()` on the `test-data` directory. How that could be clearing the directory I don't know. 🤷‍♂️

justinfagnani:
argh. I added:

```ts
  const p = execSync('ls -al test/test-data');
  console.log(p.toString());
```

to the test and it started passing. I then commented those lines out and it stayed passing. I don't get it. I'll push with those lines and see.

justinfagnani:
@aomarks that last change made the tests pass. Why was the data file being deleted before?

packages/custom-elements-manifest-tools/src/test/lib/get-custom-elements_test.ts
```diff
@@ -0,0 +1,96 @@
...
+test('Gets an element from a valid manifest, with a reference that includes the module', () => {
+  // This is the same test as before, bet the custom-element-definition reference
+  // includes the module path. This looks like a non-local reference, but it's local.
+  const newManifest = {
+    ...manifest,
+  };
+  newManifest.modules[0]!.exports![1]!.declaration.module =
+    newManifest.modules[0]!.path;
+  console.log(newManifest.modules[0]!.exports![1]!.declaration.module);
+
```
aomarks, line 73:
```suggestion
```

packages/custom-elements-manifest-tools/src/test/lib/get-custom-elements_test.ts
```diff
@@ -0,0 +1,96 @@
...
+  assert.equal(customElements.length, 1);
+  assert.equal(customElements[0]?.declaration.name, 'MyElement');
+  assert.equal(customElements[0]?.export.name, 'my-element');
+  assert.equal(customElements[0]?.module.path, 'my-project/my-element.js');
+  assert.equal(customElements[0]?.package, manifest);
+});
+
+test('Gets an element from a valid manifest, with a reference that includes the module', () => {
+  // This is the same test as before, bet the custom-element-definition reference
+  // includes the module path. This looks like a non-local reference, but it's local.
```
aomarks, line 66: Can you add a comment about why this happens? Or why there isn't always a module path?

packages/custom-elements-manifest-tools/src/test/lib/get-custom-elements_test.ts
```diff
@@ -0,0 +1,96 @@
...
+  const customElements = getCustomElements(manifest, 'test', '0.0.0');
+  assert.equal(customElements.length, 1);
+  assert.equal(customElements[0]?.declaration.name, 'MyElement');
+  assert.equal(customElements[0]?.export.name, 'my-element');
+  assert.equal(customElements[0]?.module.path, 'my-project/my-element.js');
+  assert.equal(customElements[0]?.package, manifest);
+});
+
+test('Gets an element from a valid manifest, with a reference that includes the module', () => {
+  // This is the same test as before, bet the custom-element-definition reference
```
aomarks, line 65:
```suggestion
  // This is the same test as before, but the custom-element-definition reference
```

packages/custom-elements-manifest-tools/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
     "composite": true,
+    "tsBuildInfoFile": "tsconfig.tsbuildinfo",
```
aomarks, line 5: This is already just below

aomarks (approved):

Issue #1346, opened 2022-10-03
[site-server] Add element page


justinfagnani:
We need a page to display a single element. The URL will probably be `/catalog/package/:packageName@version/element/:elementName`.

justinfagnani:
Closing as we have a very bare-bones page. We need more issues for features on the page.

PR #1349, opened 2022-10-05
Add site dev server

justinfagnani:
Adds a custom dev server based on `@web/dev-server`, along with a Koa router middleware that handles the `/catalog` URL space.

Right now there's a simple hard-coded response wired in at `/catalog/element/:path` that parses the `:path` as a `packageName/elementName` pair and shows them on the page.

Part of https://github.com/webcomponents/webcomponents.org/issues/1346

### DIFF
--- a/packages/site-server/package.json
+++ b/packages/site-server/package.json
@@ -8,7 +8,8 @@
   "type": "module",
   "scripts": {
     "build": "wireit",
-    "check": "wireit"
+    "check": "wireit",
+    "serve:dev": "wireit"
   },
   "wireit": {
     "build": {
@@ -29,6 +30,12 @@
         "src/**/*.ts"
       ],
       "output": []
+    },
+    "serve:dev": {
+      "command": "node --enable-source-maps ./dev-server.js",
+      "dependencies": [
+        "build"
+      ]
     }
   },
   "dependencies": {
@@ -36,6 +43,7 @@
     "@types/koa-conditional-get": "^2.0.0",
     "@types/koa-etag": "^3.0.0",
     "@types/koa-static": "^4.0.2",
+    "@web/dev-server": "^0.1.34",
     "koa": "^2.13.4",
     "koa-conditional-get": "^3.0.0",
     "koa-etag": "^4.0.0",
--- a/packages/site-server/src/catalog/router.ts
+++ b/packages/site-server/src/catalog/router.ts
@@ -1,0 +1,45 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import Router from '@koa/router';
+
+export const catalogRouter = new Router();
+
+catalogRouter.get('/element/:path+', async (context) => {
+  const {params} = context;
+
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  const elementPath = params['path']!;
+  const elementPathSegments = elementPath.split('/');
+  const isScoped = elementPathSegments[0]?.startsWith('@');
+  const packageName = isScoped
+    ? elementPathSegments[0] + '/' + elementPathSegments[1]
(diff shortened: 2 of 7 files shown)


packages/site-server/src/catalog/router.ts
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+import Router from '@koa/router';
+
+export const catalogRouter = new Router();
+
+catalogRouter.get('/element/:path+', async (context) => {
+  const {params} = context;
+
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
justinfagnani, line 14: I'm finding this rule to be pretty annoying and of questionable value. There are lots of places like this where the code can't reach unless this expression is non-null (in the case the pattern matching done by `.get()`, later in the file the pattern matching done by `replaceAll()`) and I'd rather not add optional-chaining or other checks just to satisfy the type-checker.

packages/site-server/package.json
```diff
@@ -29,13 +30,20 @@
         "src/**/*.ts"
       ],
       "output": []
+    },
+    "serve:dev": {
```
aomarks, line 34: Do we want this? It's already in the top-level. I kind of lean towards moving this sort of top level commands entirely to the top level, rather than doing delegation. But currently the `node` command is duplicated, so I think we want one or the other (all in top-level, or actual delegation).

packages/site-server/src/catalog/router.ts
aomarks: Sure, I don't mind removing it.

packages/site-server/src/catalog/router.ts
```diff
@@ -0,0 +1,45 @@
...
+  const elementPath = params['path']!;
+  const elementPathSegments = elementPath.split('/');
+  const isScoped = elementPathSegments[0]?.startsWith('@');
+  const packageName = isScoped
+    ? elementPathSegments[0] + '/' + elementPathSegments[1]
+    : elementPathSegments[0]!; // eslint-disable-line @typescript-eslint/no-non-null-assertion
+  const elementName = elementPathSegments[isScoped ? 2 : 1];
+  
+  context.body = `
```
aomarks, line 23: Presumably we would factor the actual route handler for each page to another module for organization? So that router.js only does routing? Fine for now though.

packages/site-server/tsconfig.json
```diff
@@ -3,8 +3,10 @@
   "compilerOptions": {
     "rootDir": "./src",
     "noEmit": true,
-    "tsBuildInfoFile": "tsconfig.tsbuildinfo"
-  },
+    "tsBuildInfoFile": "tsconfig.tsbuildinfo",
+    "module": "esnext",
```
aomarks, line 7: Do we want to standardize everything on `esnext` and `node`? I'm a little hazy on exactly what the differences are between those and `nodenext`.

aomarks (approved):
Nice! Just some small comments, pre-emptive LGTM.

packages/site-server/package.json
justinfagnani: I was going to have the command in the package be the canonical one, but before wireit service mode it's a little trickier to call. Once we have service mode I think we _do_ want them in the packages to be local, and top-level scripts that coordinate the package-level scripts.

packages/site-server/src/catalog/router.ts
justinfagnani: Yeah, the next PR moves this to a template, but puts a GraphQL call right here. Was planning on refactoring to more modules over time.

packages/site-server/tsconfig.json
justinfagnani: I guess? I think the difference has to do with package exports and ESM vs CommonJS. I just know that this works now here... not whether or for how long we should avoid the Node* settings (Node16 didn't work either).

PR #1350, opened 2022-10-05
Add rough element page

justinfagnani:
Adds a _very_ rough element page that just outputs some element fields and methods.

It looks like:

localhost_8080_catalog_element_@shoelace-style_shoelace_sl-button

There isn't much is the way of input or database value escaping or sanitization, and the page is rendered with a plain JS template string. We need to choose template system that will comprehensively sanitize data.

Part of #1346

### DIFF
--- a/packages/site-server/src/catalog/element-template.ts
+++ b/packages/site-server/src/catalog/element-template.ts
@@ -27,7 +27,6 @@ export const renderElement = ({
   manifest: Package;
 }): string => {
   const declarationRef = parseReferenceString(declarationReference);
-  console.log({declarationRef});
   if (declarationRef.module === undefined) {
     return `
       <h1>Error</h1>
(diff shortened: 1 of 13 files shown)


packages/site-server/src/catalog/element-template.ts
```diff
@@ -0,0 +1,95 @@
...
+  declarationReference: string;
+  customElementExport: string;
+  manifest: Package;
+}): string => {
+  const declarationRef = parseReferenceString(declarationReference);
+  console.log({declarationRef});
+  if (declarationRef.module === undefined) {
+    return `
+      <h1>Error</h1>
+      <p>${declarationReference} has no module</p>
```
aomarks, line 34: Should this and the other failures be a 404?

aomarks (approved):

packages/site-server/src/catalog/element-template.ts
justinfagnani: Maybe some other 4xx error? In this case the element is found, but some data is malformed. These should be checked in the improt and added to problems. I think making the template either show the errors, or be robust to bad data is the way to do.

PR #1354, opened 2022-10-07
Add Docker configurations

aomarks:
Adds Docker configurations, with a `npm run serve:docker` command that brings up a local Docker environment.

Part of https://github.com/webcomponents/webcomponents.org/issues/1339

Also:

- A new port scheme. This gives consistent unique ports to each service, in each mode, so that there are never collisions.
- Made the catalog graphql parameter a full URL, instead of just a port.
- Bumped all dependencies.


README.md
```diff
@@ -1,31 +1,52 @@
...
+
+```sh
+npm run serve:docker
+```
+
+## Ports
+
+We use the following port scheme for consistency and to prevent collisions:
+
+- [`localhost:5450`](http://localhost:5450): site-server in dev mode
```
justinfagnani, line 48: Why not 5150? (j/k bad Van Halen reference)

docker/catalog/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+# Install NPM dependencies
+COPY package.json package-lock.json ./
+COPY packages/custom-elements-manifest-tools/package.json packages/custom-elements-manifest-tools/
+COPY packages/catalog-api/package.json packages/catalog-api/
+COPY packages/catalog-server/package.json packages/catalog-server/
+RUN npm ci
+
+# Build the server
+COPY tsconfig.base.json ./
+COPY packages/custom-elements-manifest-tools/ packages/custom-elements-manifest-tools/
```
justinfagnani, line 17: why are these COPYs separate from the package.json COPYs above?

justinfagnani (approved):

docker/catalog/Dockerfile
aomarks: The first ones are just the `package.json` files, so that `npm ci` works. The second ones are the actual code, so that `npm run build` works.

The reason these are separate blocks is for efficient Docker image layering. It means that if the `package.json` files didn't change, then the `npm ci` result can be restored from cache.

Issue #1351, opened 2022-10-06
[catalog-server] Ensure DB partitioning for tests and staging purposes


justinfagnani:
We want to be able to partition the Firestore database for running multiple tests agains the same database in parallel, and to potentially support multi-tenancy, different environments in the same GCP project, etc.

To do this we need to ensure that each collection has a unique name per environment. We do this now with a namespace suffix on the `packages` collection, but this isn't quite good enough because [collection group queries](https://cloud.google.com/firestore/docs/query-data/queries#collection-group-query) could be performed on other collections. So we need to make sure that `.collection()` and `.collectionGroup()` are never called directly, but called through utility methods that add the namespace suffix.

We also need to update the design doc to describe this pattern.

justinfagnani:
There's a problem with expressing namespace as collection suffixes: indices. Indices must be created via the console or CLI, and take a few minutes to setup even on empty databases. Indices are also configured by collection IDs, and are _required_ for collection group queries.

The other approach we can take is to add a `namespace` field to every document and always filter on that field.

justinfagnani:
This wasn't fully fixed yet

PR #1359, opened 2022-10-08
Implement basic full-text-search

justinfagnani:
This implements full-text search by:
1. Stemming a number of fields and storing the results in an array field in CustomElement documents
2. Stemming the user's text query
3. Retrieving elements with a collectionGroup query across the `customElement` collections and filtering with an `array-contains-any` operator

I'm using https://github.com/NaturalNode/natural for stemming. It looks well maintained and thorough compared to other libraries.

The fields currently indexed are:
- package description
- element summary
- element description
- A split of the element name by `-`, using segments more than 3 characters long (to get `button` from `sl-button`, etc).

Builds on https://github.com/webcomponents/webcomponents.org/pull/1364. We need to fix https://github.com/webcomponents/webcomponents.org/issues/1351 to ensure that queries are confined to a single namespace.

### DIFF
--- a/packages/catalog-api/src/lib/schema.graphql
+++ b/packages/catalog-api/src/lib/schema.graphql
@@ -8,27 +8,8 @@ type Query {
   """
   Queries custom elements in the entire catalog, from the latest version of each
   package.
-
-  Eventually this will have more query parameters, and use some sort of ranking
-  algorithm, otherwise the order will just be defined by the database
-  implementation.
-
-  NOT IMPLEMENTED
-  DO_NOT_LAUNCH
   """
-  elements(distTag: String = "latest", limit: Int): [CustomElement!]!
-
-  """
-  Retrieves the custom element data for a single element.
-
-  NOT IMPLEMENTED
-  DO_NOT_LAUNCH
-  """
-  element(
-    packageName: String!
-    elementName: String!
-    tag: String = "latest"
-  ): CustomElement
+  elements(query: String, limit: Int): [CustomElement!]!
 }
 
 type Mutation {
--- a/packages/catalog-server/src/lib/catalog.ts
+++ b/packages/catalog-server/src/lib/catalog.ts
@@ -272,8 +272,7 @@ export class Catalog {
 
     console.log('Writing custom elements...');
     await this.#repository.writeCustomElements(
-      packageName,
-      version,
+      packageVersionMetadata,
       customElements,
       versionDistTags,
       author
@@ -307,11 +306,21 @@ export class Catalog {
     return this.#repository.getPackageVersion(packageName, version);
   }
 
+  /**
+   * Gets the custom elements for a package
+   */
   async getCustomElements(
     packageName: string,
     version: string,
     tagName: string | undefined
   ): Promise<Array<CustomElement>> {
     return this.#repository.getCustomElements(packageName, version, tagName);
+  }
+
+  async queryElements({query, limit}: {query?: string; limit?: number}) {
+    // TODO (justinfagnani): The catalog should parse out GitHub-style search
+    // operators (like "author:yogibear") and pass structured + text queries
+    // to the repository
+    return this.#repository.queryElements({query, limit});
   }
 }
--- a/packages/catalog-server/src/lib/firestore/firestore-repository.ts
+++ b/packages/catalog-server/src/lib/firestore/firestore-repository.ts
@@ -160,16 +160,22 @@ export class FirestoreRepository implements Repository {
           // collection first.
           const elements = await t.get(customElementsRef);
 
+          const isLatest = versionDistTags.includes('latest');
+
           // Update the PackageVersion doc
-          await t.update(versionRef, {
+          // We remove the converter to fix the types:
+          // https://github.com/googleapis/nodejs-firestore/issues/1745
+          await t.update(versionRef.withConverter(null), {
             distTags: versionDistTags,
+            isLatest,
           });
 
           // Update all elements
           await Promise.all(
             elements.docs.map(async (element) => {
-              await t.update(element.ref, {
+              await t.update(element.ref.withConverter(null), {
                 distTags: versionDistTags,
+                isLatest,
               });
             })
           );
@@ -284,6 +290,7 @@ export class FirestoreRepository implements Repository {
     // Store custom elements data in subcollection
     const versionRef = this.getPackageVersionRef(packageName, version);
     const customElementsRef = versionRef.collection('customElements');
+    const isLatest = distTags.includes('latest');
     const batch = db.batch();
 
     for (const c of customElements) {
@@ -291,6 +298,7 @@ export class FirestoreRepository implements Repository {
         package: packageName,
         version,
         distTags,
+        isLatest,
         author,
         tagName: c.export.name,
         className: c.declaration.name,
@@ -394,10 +402,14 @@ export class FirestoreRepository implements Repository {
       }
 
       // Now query for a version that's assigned this dist-tag
-      const result = await this.getPackageVersionCollectionRef(packageName)
-        .where('distTags', 'array-contains', versionOrTag)
-        .limit(1)
-        .get();
+      let query: CollectionReference<PackageVersion> | Query<PackageVersion> =
+        this.getPackageVersionCollectionRef(packageName);
+      if (versionOrTag === 'latest') {
+        query = query.where('isLatest', '==', true);
+      } else {
+        query = query.where('distTags', 'array-contains', versionOrTag);
+      }
+      const result = await query.limit(1).get();
       if (result.size !== 0) {
         return result.docs[0]!.data();
       }
--- a/packages/catalog-server/src/lib/firestore/firestore-repository.ts
+++ b/packages/catalog-server/src/lib/firestore/firestore-repository.ts
@@ -511,7 +511,6 @@ export class FirestoreRepository implements Repository {
       if (queryTerms.length > 10) {
         queryTerms.length = 10;
       }
-      console.log('queryTerms', queryTerms);
       dbQuery = dbQuery.where('searchTerms', 'array-contains-any', queryTerms);
     }
 
--- a/packages/catalog-server/src/lib/firestore/firestore-repository.ts
+++ b/packages/catalog-server/src/lib/firestore/firestore-repository.ts
@@ -325,7 +325,10 @@ export class FirestoreRepository implements Repository {
           ...descriptionStems,
           ...summaryStems,
           ...tagNameParts,
-          tagName
+          tagName,
+          // TODO (justinfagnani): tokenizing the package name is temporary
+          // until we don't tokenize the *entire* query
+          ...natural.PorterStemmer.tokenizeAndStem(packageName),
         ]),
       ];
 
@@ -508,6 +511,7 @@ export class FirestoreRepository implements Repository {
       if (queryTerms.length > 10) {
         queryTerms.length = 10;
       }
+      console.log('queryTerms', queryTerms);
       dbQuery = dbQuery.where('searchTerms', 'array-contains-any', queryTerms);
     }
 
--- a/packages/catalog-server/src/lib/firestore/firestore-repository.ts
+++ b/packages/catalog-server/src/lib/firestore/firestore-repository.ts
@@ -4,13 +4,19 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import {FieldValue, Query, CollectionReference} from '@google-cloud/firestore';
+import {
+  FieldValue,
+  Query,
+  CollectionReference,
+  CollectionGroup,
+} from '@google-cloud/firestore';
 import {Firestore} from '@google-cloud/firestore';
 import firebase from 'firebase-admin';
 import {CustomElementInfo} from '@webcomponents/custom-elements-manifest-tools';
 import {referenceString} from '@webcomponents/custom-elements-manifest-tools/lib/reference-string.js';
 import clean from 'semver/functions/clean.js';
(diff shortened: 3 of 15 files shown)


packages/catalog-api/src/lib/schema.graphql
```diff
@@ -8,27 +8,8 @@ type Query {
...
-
-  NOT IMPLEMENTED
-  DO_NOT_LAUNCH
   """
-  element(
-    packageName: String!
-    elementName: String!
-    tag: String = "latest"
-  ): CustomElement
+  elements(query: String, limit: Int): [CustomElement!]!
```
aomarks, line 12: Do we perhaps also want a strict lookup like the `element` API we had before, or will we instead support search prefixes like `package:foo element:bar`?

packages/catalog-server/src/lib/catalog.ts
```diff
@@ -307,11 +306,21 @@ export class Catalog {
...
   async getCustomElements(
     packageName: string,
     version: string,
     tagName: string | undefined
   ): Promise<Array<CustomElement>> {
     return this.#repository.getCustomElements(packageName, version, tagName);
   }
+
+  async queryElements({query, limit}: {query?: string; limit?: number}) {
+    // TODO (justinfagnani): The catalog should parse out GitHub-style search
```
aomarks, line 321: Ah, answers my question above I guess. Mind filing an issue?

packages/catalog-server/src/lib/firestore/firestore-repository.ts
```diff
@@ -160,16 +169,22 @@ export class FirestoreRepository implements Repository {
           // collection first.
           const elements = await t.get(customElementsRef);
 
+          const isLatest = versionDistTags.includes('latest');
+
           // Update the PackageVersion doc
-          await t.update(versionRef, {
+          // We remove the converter to fix the types:
```
aomarks, line 175: Not sure I understand "remove the converter" here, since it's actually adding "withConverter" -- seems like the opposite? :)

packages/catalog-server/src/lib/firestore/firestore-repository.ts
```diff
@@ -275,31 +290,60 @@ export class FirestoreRepository implements Repository {
...
   async writeCustomElements(
-    packageName: string,
-    version: string,
+    packageVersionMetadata: Version,
     customElements: CustomElementInfo[],
     distTags: string[],
     author: string
   ): Promise<void> {
     // Store custom elements data in subcollection
+    const {name: packageName, version, description} = packageVersionMetadata;
```
aomarks, line 299: Nit: looks like you could destructure this in the function parameter list?

packages/catalog-server/src/lib/firestore/firestore-repository.ts
```diff
@@ -275,31 +290,60 @@ export class FirestoreRepository implements Repository {
...
+    // Stem the package description
+    const packageDescriptionStems = natural.PorterStemmer.tokenizeAndStem(
+      description ?? ''
+    );
+
     for (const c of customElements) {
+      const tagName = c.export.name;
+      // Grab longer tag name parts for searching. We want "button" from
+      // md-button, etc.
+      const tagNameParts = tagName.split('-').filter((s) => s.length > 3);
```
aomarks, line 314: Maybe even 2? Thinking of `md-` which I heard might be the new prefix for mwc?

packages/catalog-server/src/lib/firestore/firestore-repository.ts
```diff
@@ -275,31 +290,60 @@ export class FirestoreRepository implements Repository {
...
+        c.declaration.summary ?? ''
+      );
+
+      // Combine and deduplicate terms
+      const searchTerms = [
+        ...new Set([
+          ...packageDescriptionStems,
+          ...descriptionStems,
+          ...summaryStems,
+          ...tagNameParts,
```
aomarks, line 328: Just an intuition, no action needed. Feels like we might want to weight tag name parts a little more highly than description terms?

packages/catalog-server/src/lib/firestore/firestore-repository.ts
aomarks, line 314: Should we also directly include the un-split tag name as a separate term? It feels like that could naturally boost an exact search someone is trying, but without a "tag:" prefix, like "mwc-button".

packages/catalog-server/src/lib/firestore/firestore-repository.ts
```diff
@@ -441,6 +489,32 @@ export class FirestoreRepository implements Repository {
...
+      .collectionGroup('customElements')
+      .withConverter(customElementConverter)
+      .where('isLatest', '==', true)
+      .limit(limit ?? 25);
+
+    if (query !== undefined) {
+      // Split query
+      const queryTerms = natural.PorterStemmer.tokenizeAndStem(query);
+      if (queryTerms.length > 10) {
+        queryTerms.length = 10;
```
aomarks, line 509: Why limit at 10?

packages/catalog-server/src/lib/firestore/firestore-repository.ts
```diff
@@ -441,6 +489,32 @@ export class FirestoreRepository implements Repository {
...
+      }
+      dbQuery = dbQuery.where('searchTerms', 'array-contains-any', queryTerms);
```
aomarks, line 511: We discussed offline, but just for the record here -- it sounds like the `searchTerms` field is automatically indexed, and firestore supports indexed `array-contains-any` lookups.

packages/catalog-server/src/lib/graphql.ts
```diff
@@ -53,6 +62,13 @@ export const makeExecutableCatalogSchema = async (catalog: Catalog) => {
           return null;
         }
       },
+      async elements(_parent, {query, limit}): Promise<Array<CustomElement>> {
+        console.log('query elements', {query, limit});
```
aomarks, line 66: Remove log?

packages/catalog-server/src/lib/firestore/firestore-repository.ts
```diff
@@ -441,6 +489,32 @@ export class FirestoreRepository implements Repository {
     return customElementsResults.docs.map((d) => d.data());
   }
 
+  async queryElements({
+    query,
+    limit,
+  }: {
+    query?: string;
+    limit?: number;
```
aomarks, line 497: Can limit be non-optional here, since the same `25` default is assigned earlier in the call stack? Or maybe factored out to a constant? (Just thinking if we want to change the default, it would be easy to only do it in one place).

packages/catalog-server/src/test/lib/catalog_test.ts
```diff
@@ -25,6 +25,13 @@ const testPackage1Path = fileURLToPath(
   new URL('../test-packages/test-1/', import.meta.url)
 );
 
+// A set of import, fetch, search tests that use the same data
+const TEST_SEQUENCE_ONE = 'test-data-1';
```
aomarks, line 29: Slightly confused by the term "sequence" here, it's actually a partition name? Or is sequence some technical firestore term?

packages/catalog-server/src/lib/firestore/firestore-repository.ts
```diff
@@ -441,6 +489,32 @@ export class FirestoreRepository implements Repository {
...
+
+    if (query !== undefined) {
+      // Split query
+      const queryTerms = natural.PorterStemmer.tokenizeAndStem(query);
+      if (queryTerms.length > 10) {
+        queryTerms.length = 10;
+      }
+      dbQuery = dbQuery.where('searchTerms', 'array-contains-any', queryTerms);
+    }
+
```
aomarks, line 513: File an issue for ranking and add TODO? As discussed offline, I think a word frequency table (computed from our own data) with a https://en.wikipedia.org/wiki/Tf%E2%80%93idf score would be a really simple and good start.

aomarks (changes requested):
Nice! I can see us moving to something e.g. Lucene based, or a separate hosted service in the future -- but I like how simple this implementation is and am happy to see how far it can take us (especially after adding some ranking).

Main comment is for more test coverage :D

packages/catalog-server/src/test/lib/catalog_test.ts
```diff
@@ -84,6 +91,37 @@ test('A second import does nothing', async () => {
   assert.equal(importResult.problems, undefined);
 });
 
+test('Full text search', async () => {
```
aomarks, line 94: (I thought I sent this comment with the review, but must have lost it somehow).

Feels like we could do with some more test coverage here. Maybe a test rig/utility function that would make it easier to concisely write the scenarios? Some cases it feels like we should have:

- Search by tag name term
- Search by package term
- Search by entire package name
- Search by entire tag name
- Search by description term
- Tests for the various stemmings/other normalization we expect to be happening -- case sensitivity, removing punctuation, plural normalization, etc.
- A search with both a matching and non-matching term (to confirm it's OR not AND matching)
- Checking that the limit is respected
- No matching results
- No elements in the index at all

aomarks (changes requested):

packages/catalog-api/src/lib/schema.graphql
justinfagnani: I think that the `package()` query gives us what we had with `element()` - and we can make some of the fields like `customElements` into queries if we want that.

I think this query should grow to add both search operators and possibly subqueries/fields that get you from element to package here.

One consideration is how to implement structured + text search. If we move to use a search service do we do both Firestore + search service queries and mix them? If a search service offer structured fields do we delegate all searches to it?

packages/catalog-server/src/lib/catalog.ts
justinfagnani: https://github.com/webcomponents/webcomponents.org/issues/1366

packages/catalog-server/src/lib/firestore/firestore-repository.ts
justinfagnani: I think weighting / relevance might be one reason to push us to use a search service rather quickly. I don't think we have a natural way of getting Firestore to consider some kind of weighting in our query, we just get a result or not for the `array-contains-any` operator. Sorting by array fields [isn't very useful](https://cloud.google.com/firestore/docs/concepts/data-types). We can get Firestore to sort by simple fields like a "ranking" field.

To do a relevance ordering we'd have to talk all the results, iterate and apply something like tf-id, and sort on that. That means our Firestore results size will have to be bigger than our requested query size because  we might have high-relevance results later in the Firestore result set.

It's plausible we could continue down this path and keep it simple, but it also seems like it might get complicated real quick.

packages/catalog-server/src/lib/firestore/firestore-repository.ts
justinfagnani: It actually seems fine for it to differ. The graphql one is what matters most. This is just a fallback incase someone called the repo directly without a limit. Required is fine too, though in a JS world I'd still usually check before launching a non-limited query :)

packages/catalog-server/src/lib/firestore/firestore-repository.ts
justinfagnani: That's the Firestore limit: https://cloud.google.com/firestore/docs/query-data/queries#array-contains-any

I should make a TODO/issue - If we keep this approach, we want to sort the query by inverse frequency so we keep the most important terms. Then 10 seems like enough. For longer queries I _think_ you get better results with n-gram searches, but that's stretching my knowledge and probably another reason to use a service :)

packages/catalog-server/src/lib/firestore/firestore-repository.ts
justinfagnani: Yep, and there are no errors in the emulator.

packages/catalog-server/src/lib/firestore/firestore-repository.ts
justinfagnani: https://github.com/webcomponents/webcomponents.org/issues/1361

packages/catalog-server/src/lib/graphql.ts
justinfagnani: done

packages/catalog-server/src/test/lib/catalog_test.ts
justinfagnani: It's my term... I'm meaning a sequence of tests that rely on each other. I really want some kind of test nesting.

packages/catalog-server/src/lib/firestore/firestore-repository.ts
aomarks: Good point,it  didn't occur to me that we'd need the ranking to happen on the database side, and presumably you couldn't write a query like that (pretty sure you could with something like postgres, but firebase doubtful). I'm pretty convinced we'll need a standalone search service, then, based on lucene or similar (a service we run, or a third-party hosted one).

packages/catalog-server/src/test/lib/catalog_test.ts
justinfagnani: Agree we should definitely have more coverage. I added a few more tests for now.

packages/catalog-server/src/lib/firestore/firestore-repository.ts
```diff
@@ -508,6 +511,7 @@ export class FirestoreRepository implements Repository {
       if (queryTerms.length > 10) {
         queryTerms.length = 10;
       }
+      console.log('queryTerms', queryTerms);
```
aomarks, line 514: Remove log

aomarks (approved):

PR #1360, opened 2022-10-12
Fix bugs with updating packages

justinfagnani:
The main test added sets up the local project so that it has a new published version, then performs an import, to test updating a package. This revealed several bugs that are fixed in the commit too.

To support this type of test I updated LocalFsPackageFiles to point to a folder of versioned packaged data, and to take config of which versions are considered publish and what the dist-tags are. That's in the first commit.


packages/catalog-server/src/lib/catalog.ts
```diff
@@ -154,6 +154,12 @@ export class Catalog {
         }
       }
 
+      // Remove the version to import from versionsToUdpate, because it
```
aomarks, line 157: versionsToUpdate

aomarks (approved):

PR #1365, opened 2022-10-13
Add base template for 11ty and catalog pages

justinfagnani:


packages/site-content/package.json
```diff
@@ -15,7 +15,8 @@
       "command": "eleventy --config=eleventy.config.prod.cjs",
       "files": [
         "eleventy.config.prod.cjs",
-        "site"
+        "site",
+        "../site-server/src/templates/base.ts"
```
aomarks, line 19: I think this should be the `.js` file, not the `.ts` file, otherwise there is a race condition.

packages/site-content/package.json
```diff
@@ -29,14 +30,18 @@
       "files": [
         "eleventy.config.dev.cjs",
         "site",
-        "!site/assets"
+        "!site/assets",
+        "../site-server/src/templates/base.ts"
```
aomarks, line 34: `.js` not `.ts` (see above)

packages/site-content/package.json
```diff
@@ -29,14 +30,18 @@
...
       ],
       "output": [
         "_dev"
       ]
-    }
+    }
```
aomarks, line 39:
```suggestion
    }
```

packages/site-server/src/templates/base.ts
```diff
@@ -0,0 +1,52 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import {escapeHTML} from '../catalog/escape-html.js';
+
+export const renderPage = (data: {
+  scripts?: Array<string>;
+  title?: string;
+  content?: string;
+}) => {
+  return `<!doctype html>
```
aomarks, line 13: WDYT about a no-op html tag function so that we get syntax highlighting?

packages/site-content/site/index.html
```diff
@@ -1,17 +1,8 @@
...
-    <script type="module" src="./js/homepage.js"></script>
-  </body>
-</html>
+---
+layout: "layouts/base.11ty.cjs"
+title: "Web Components"
+scripts:
+- "./js/homepage.js"
+---
+<wco-top-bar></wco-top-bar>
```
aomarks, line 7: So this file is still using nunjucks, but the result goes to the `content` parameter.

I guess I feel slightly fuzzy about why this page is a `html` file, but the base template is `js`. Is it possible for us to do `js` all the way down?

aomarks (changes requested):
Cool!

packages/site-content/package.json
aomarks, line 19: Maybe `../site-server/lib/templates/**/*.js` so that we don't need to enumerate if we add more?

aomarks (changes requested):

packages/site-content/site/index.html
justinfagnani: Every layout/template can be in a different language in 11ty. We could make the content page in JS, but I think that HTML or markdown + Nunjucks is a pretty good experience for content pages. The layout is in JS instead of Nunjucks mainly to make it easy to call in the server-generated pages without running it through a template system.

packages/site-server/src/templates/base.ts
justinfagnani: Seems fine, though I might wait until we integrate lit-ssr and see if we just use that everywhere.

aomarks (approved):

PR #1370, opened 2022-10-17
Add a stub search UI

justinfagnani:
This is a very basic UI with one search field.

It performs a client-side GraphQL query. I used urql on the client side instead of Apollo because urql is smaller and published as standard JS modules (except for some `process.env` calls). The integration point is quite small and can be swapped out. If we keep it we need to move the client to a common location and probably want to wrap it in a controller.

There are a bunch of rough edges here intentionally. To make sharable links and edit-refresh development easier I added a `?query=` query param, but didn't yet use a router to read and write it. The styling is made to be obviously sketch-quality, but I added an "icon" to hint at what kind of presentation we might want.

Being able to show links to all elements imported in the bootstrap import uncovers some bugs - many elements display an error on their page. The Shoelace components work though.

Here's a preview:

localhost_5450_catalog_query=button

### DIFF
--- a/packages/site-client/package.json
+++ b/packages/site-client/package.json
@@ -23,7 +23,7 @@
       ]
     },
     "build:dev": {
-      "command": "find src -name \"*.ts\" | xargs esbuild --format=esm --target=es2022 --color --outdir=lib",
+      "command": "find src -name \"*.ts\" | xargs esbuild --color --outdir=lib",
       "files": [
         "src/**/*.ts"
       ],
@@ -33,7 +33,7 @@
       "clean": "if-file-deleted"
     },
     "build:prod": {
-      "command": "esbuild --format=esm --target=es2022 --color --outdir=bundled --bundle --splitting --minify src/entrypoints/*.ts",
+      "command": "esbuild --color --outdir=bundled --bundle --splitting --minify src/entrypoints/*.ts",
       "files": [
         "src/**/*.ts"
       ],
@@ -66,6 +66,9 @@
     }
   },
   "dependencies": {
+    "@urql/core": "^3.0.4",
+    "@webcomponents/catalog-api": "^0.0.0",
+    "graphql": "^16.6.0",
     "lit": "^2.3.1",
     "lit-analyzer": "^1.2.1"
   }
--- a/packages/site-client/src/components/wco-catalog-search.ts
+++ b/packages/site-client/src/components/wco-catalog-search.ts
@@ -4,28 +4,87 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
- import {html, css, LitElement} from 'lit';
- import {customElement} from 'lit/decorators.js';
- 
- @customElement('wco-catalog-search')
- export class WCOCatalogSearch extends LitElement {
-   static styles = css`
-     :host {
-       display: flex;
-       align-items: center;
-     }
-   `;
- 
-   render() {
-     return html`
-       Search: <input>
-     `;
-   }
- }
- 
- declare global {
-   interface HTMLElementTagNameMap {
-     'wco-catalog-search': WCOCatalogSearch;
-   }
- }
- 
+import {html, css, LitElement} from 'lit';
+import {customElement, query, state} from 'lit/decorators.js';
+import {gql, createClient, defaultExchanges} from '@urql/core';
+import type {CustomElement} from '@webcomponents/catalog-api/lib/schema.js';
+
+import './wco-element-card.js';
+
+const client = createClient({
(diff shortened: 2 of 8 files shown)


packages/site-client/package.json
```diff
@@ -33,7 +33,7 @@
       "clean": "if-file-deleted"
     },
     "build:prod": {
-      "command": "esbuild --format=esm --target=es2022 --color --outdir=bundled --bundle --splitting --minify src/entrypoints/*.ts",
+      "command": "esbuild --color --outdir=bundled --bundle --splitting --minify src/entrypoints/*.ts",
```
aomarks, line 36: Format is picked automatically when bundling is enabled, but I couldn't figure out from the docs which format it's supposed to be: https://esbuild.github.io/api/#format

Here and above, why not keep the explicit `--format=esm` flag?

packages/site-client/package.json
```diff
@@ -23,7 +23,7 @@
       ]
     },
     "build:dev": {
-      "command": "find src -name \"*.ts\" | xargs esbuild --format=esm --target=es2022 --color --outdir=lib",
+      "command": "find src -name \"*.ts\" | xargs esbuild --color --outdir=lib",
```
aomarks, line 26: The default is `--target=esnext` -- that doesn't seem right, even for dev?

packages/site-client/src/components/wco-catalog-search.ts
```diff
@@ -4,28 +4,87 @@
...
- declare global {
-   interface HTMLElementTagNameMap {
-     'wco-catalog-search': WCOCatalogSearch;
-   }
- }
- 
\ No newline at end of file
+import {html, css, LitElement} from 'lit';
+import {customElement, query, state} from 'lit/decorators.js';
+import {gql, createClient, defaultExchanges} from '@urql/core';
```
aomarks, line 9: I thought we were using apollo, why are we also using urql?

packages/site-client/src/components/wco-element-card.ts
```diff
@@ -0,0 +1,99 @@
...
+  private element?: CustomElement;
+
+  render() {
+    if (this.element === undefined) {
+      return undefined;
+    }
+    const packageName = this.element.package;
+    const tagName = this.element.tagName ?? '';
+
+    // Generate one or two-letter initials starting after what we assyme is
```
aomarks, line 77:
```suggestion
    // Generate one or two-letter initials starting after what we assume is
```

packages/site-client/tsconfig.json
```diff
@@ -3,7 +3,8 @@
   "compilerOptions": {
     "rootDir": "./src",
     "noEmit": true,
-    "tsBuildInfoFile": "tsconfig.tsbuildinfo"
+    "tsBuildInfoFile": "tsconfig.tsbuildinfo",
+    "useDefineForClassFields": false
```
aomarks, line 7: This doesn't actually do anything since we don't emit from tsc, right? Or does esbuild automatically read this file?

packages/site-server/src/templates/base.ts
```diff
@@ -23,6 +23,11 @@ export const renderPage = (data: {
       href="https://fonts.googleapis.com/css?family=Material+Icons&display=block"
       rel="stylesheet"
     />
+    <!-- TODO (justinfagnani): only add this in dev. In prod we should have
```
aomarks, line 26: Which library needs this?

aomarks (changes requested):

packages/site-client/package.json
justinfagnani: If you specify these flags then esbuild doesn't read the tsconfig and it sets `useDefineForClassFields` to `true`, which breaks our elements.

packages/site-client/package.json
justinfagnani: We have this set in the tsconfig

packages/site-client/src/components/wco-catalog-search.ts
justinfagnani: I wanted to try it out, and it's pure modules so doesn't require the common JS plug-in in web dev server.

packages/site-client/tsconfig.json
justinfagnani: esbuild reads it now without the flags I removed.

packages/site-server/src/templates/base.ts
justinfagnani: The graphql library that urql depends on.

aomarks (approved):

PR #1372, opened 2022-10-20
Fix bug with storing element declaration references

### LINKED ISSUE #1371: [catalog-server] Element declarations in separate modules not found
Try `@lion/button/lion-button-submit` in a local dev environment: http://localhost:5450/catalog/element/@lion/button/lion-button-submit

This page errors because no declaration is found.

There are at least three problems here:
1. The page should show more about the error
2. No problems are recorded with the PackageVersion document
3. The manifest appears correct. The custom element definition and _declaration_ (the class) are in separate modules and we don't seem to resolve the reference to the declaration correctly.

We see the custom element export in the `lion-button-submit.js` module:

```json
    {
      "kind": "javascript-module",
      "path": "lion-button-submit.js",
      "declarations": [],
      "exports": [
        {
          "kind": "custom-element-definition",
          "name": "lion-button-submit",
          "declaration": {
            "name": "LionButtonSubmit",
            "module": "/src/LionButtonSubmit.js"
          }
        }
      ]
    },
```

The module has no declarations because it just imports and registers an element:
```js
import { LionButtonSubmit } from './src/LionButtonSubmit.js';

customElements.define('lion-button-submit', LionButtonSubmit);
```

The declaration is in the `src/LionButtonSubmit.js` file:

```json
    {
      "kind": "javascript-module",
      "path": "src/LionButtonSubmit.js",
      "declarations": [
        {
          "kind": "class",
          "description": ...,
          "name": "LionButtonSubmit",
```

justinfagnani:
The FirestoreRepository was calculating the reference string for the element _declaration_ incorrectly when the definition/export (the define() call) was in a different module from the declaration (the class). It was assuming that the modules were the same, so the element page couldn't then find the declaration in the manifest when it was showing element details.

This adds a few tests that were passing before the fix as I investigated the issue, and one catalog test that fails without the fix.

Fixes #1371

### DIFF
--- a/packages/site-server/src/catalog/routes/element/element-template.ts
+++ b/packages/site-server/src/catalog/routes/element/element-template.ts
@@ -51,7 +51,11 @@ export const renderElement = ({
   console.log('elementDeclaration', declaration);
 
   if (declaration === undefined || declaration.kind !== 'class') {
-    return `<h1>Error</h1>`;
+    console.log('tried to resolve', module.path, declarationRef);
+    return `
+      <h1>Error</h1>
+      <h2>Element declaration not found</h2>
+    `;
   }
 
   return `
--- a/packages/site-server/src/catalog/routes/element/element-template.ts
+++ b/packages/site-server/src/catalog/routes/element/element-template.ts
@@ -51,7 +51,6 @@ export const renderElement = ({
   console.log('elementDeclaration', declaration);
 
   if (declaration === undefined || declaration.kind !== 'class') {
-    console.log('tried to resolve', module.path, declarationRef);
     return `
       <h1>Error</h1>
       <h2>Element declaration not found</h2>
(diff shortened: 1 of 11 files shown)


packages/site-server/src/catalog/routes/element/element-template.ts
```diff
@@ -51,7 +51,11 @@ export const renderElement = ({
   console.log('elementDeclaration', declaration);
 
   if (declaration === undefined || declaration.kind !== 'class') {
-    return `<h1>Error</h1>`;
+    console.log('tried to resolve', module.path, declarationRef);
```
aomarks, line 54: remove log?

aomarks (approved):

packages/site-server/src/catalog/routes/element/element-template.ts
justinfagnani: done

Issue #1338, opened 2022-09-28
[site-server] Integrate @lit-labs/ssr


aomarks:

justinfagnani:
I think we can close then now as we have SSR + hydration w/ data on the catalog pages.

PR #1375, opened 2022-10-20
Use Lit SSR to render the element page

justinfagnani:
SSRs the element page.

I had to change the common base template to a generator to get streaming, and change 11ty wrapper to join the results of the generator. Other than that it was a very straightforward addition.

There's no top-bar on the element page right now because the JS isn't loaded for that. I'll fix in a follow-up.

Part of #1338


aomarks (approved):
Nice! Pretty simple.

PR #1381, opened 2022-10-21
Add catalog page component

justinfagnani:
This puts all of the catalog page into a component, making it easy to associate styles with the page.

It also fixes the build error in site-client and adds a script to run the prod site-server. I used that to check that reactive properties still work when using the `--format=esm` flag on esbuild.

### DIFF
--- a/packages/site-client/src/components/wco-catalog-page.ts
+++ b/packages/site-client/src/components/wco-catalog-page.ts
@@ -1,0 +1,33 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {html, css, LitElement} from 'lit';
+import {customElement} from 'lit/decorators.js';
+import './wco-top-bar.js';
+import './wco-catalog-search.js';
+
+@customElement('wco-catalog-page')
+export class WCOCatalogPage extends LitElement {
+  static styles = css`
+    :host {
+      display: block;
+    }
+  `;
+
+  render() {
+    return html`
+      <wco-top-bar></wco-top-bar>
+      <h1>Catalog</h1>
+      <wco-catalog-search></wco-catalog-search>
+    `;
+  }
+}
+
+declare global {
+  interface HTMLElementTagNameMap {
+    'wco-catalog-page': WCOCatalogPage;
+  }
+}
--- a/packages/site-server/package.json
+++ b/packages/site-server/package.json
@@ -9,7 +9,8 @@
   "scripts": {
     "build": "wireit",
     "check": "wireit",
-    "serve:dev": "wireit"
+    "serve:dev": "wireit",
+    "serve:prod": "wireit"
   },
   "wireit": {
     "build": {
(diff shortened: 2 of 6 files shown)


packages/site-server/package.json
```diff
@@ -9,13 +9,15 @@
   "scripts": {
     "build": "wireit",
     "check": "wireit",
-    "serve:dev": "wireit"
+    "serve:dev": "wireit",
```
aomarks, line 12: `start` here and below?

packages/site-server/package.json
```diff
@@ -39,6 +41,12 @@
       "dependencies": [
         "build"
       ]
+    },
+    "serve:prod": {
+      "command": "node --enable-source-maps ./lib/prod-server.js",
```
aomarks, line 46: I think we already have this script in the top-level. I'm ok with it being here instead, but I think we should remove it from the top-level then.

packages/site-client/src/components/wco-catalog-page.ts
```diff
@@ -0,0 +1,34 @@
...
+     `;
+   }
+ }
+ 
+ declare global {
+   interface HTMLElementTagNameMap {
+     'wco-catalog-page': WCOCatalogPage;
+   }
+ }
+
```
aomarks, line 34: Very nit, but files should generally end in newlines, I think? I always enable "Insert Final Newline" in vscode. Maybe just an outdated habit, but not including it can make certain shell operations a bit weird, e.g. if you `cat` two files together, the last line of the first file will be prepended to the first line of the second file.

packages/site-server/package.json
```diff
@@ -9,13 +9,15 @@
...
+    "serve:dev": "wireit",
+    "serve:prod": "wireit"
   },
   "wireit": {
     "build": {
       "command": "find src -name \"*.ts\" | xargs esbuild --format=esm --target=es2022 --color --outdir=lib",
       "dependencies": [
-        "../custom-elements-manifest-tools:build"
+        "../custom-elements-manifest-tools:build",
+        "../site-client:build"
```
aomarks, line 20: I think since we don't have "soft" dependencies in wireit yet, I would move this to the `serve:prod` and `serve:dev` scripts. Otherwise in service mode, we'll restart the server whenever the assets change, which we don't want.

aomarks (changes requested):

packages/site-server/package.json
justinfagnani: Sorry, should have pointed out, these were stacked PRs and the rename one is on top of this one.

packages/site-server/package.json
justinfagnani: k

packages/site-client/src/components/wco-catalog-page.ts
justinfagnani: Oops, I hadn't formatted and the whole file has an extra leading space for some reason.

packages/site-server/package.json
justinfagnani: We do need that now because these are used in the server.

aomarks (approved):

PR #1383, opened 2022-10-21
Add <wco-element-page>

justinfagnani:
Like #1381 but for the element route.

To get this one to work I had to disable hydration because we don't serialize initial data and send it to the client yet.


packages/site-server/src/catalog/routes/element/element-route.ts
```diff
@@ -78,22 +82,26 @@ export const handleElementRoute = async (
...
   context.status = 200;
+  context.body = Readable.from(
+    renderPage({
+      title: `${packageName}/${elementName}`,
+      scripts: ['/js/hydrate.js', '/js/element.js'],
+      // TODO (justinfagnani): If we want to hydrate, we need to serialize the
+      // initial data and embed in the response
+      content: render(
+        html`<wco-element-page .elementData=${elementData}></wco-element-page>`,
+        {deferHydration: true}
```
justinfagnani, line 103: @augustjk @kevinpschaaf @aomarks SSR DX issue here: the error I got before adding `deferHydration: true` was that an unexpected TemplateResult was encountered on the client (due to a conditional in the element when no data was present). That would have been hard to track down for a non-team-member.

aomarks (approved):

PR #1384, opened 2022-10-21
Hydrate the element page

justinfagnani:
This isn't strictly necessary yet, but will be as soon as we have interactive custom elements on the page.

I did a pretty manual thing here and serialized the data as JSON at the end of the response, then added another entrypoint that calls `hydrate()` with the data and page template.

cc @kevinpschaaf @augustjk for SSR patterns

### DIFF
--- a/packages/site-server/src/templates/base.ts
+++ b/packages/site-server/src/templates/base.ts
@@ -9,6 +9,8 @@ export function* renderPage(data: {
   scripts?: Array<string>;
   title?: string;
   content: string | Iterable<string>;
+  initialData?: object;
+  initScript?: string;
 }) {
   yield `<!doctype html>
 <html lang="en">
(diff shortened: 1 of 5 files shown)


aomarks (approved):

packages/site-server/src/templates/base.ts
```diff
@@ -58,6 +60,17 @@ export function* renderPage(data: {
   } else {
     yield* data.content;
   }
+
+  if (data.initialData !== undefined) {
+    yield `<script>window.__ssrData = ${JSON.stringify(
+      data.initialData
+    )}</script>`;
```
kevinpschaaf, line 67: I think it's good practice to sanitize from HTML injection this here (this is what Redux [suggests](https://redux.js.org/usage/server-rendering#inject-initial-component-html-and-state) in their docs):
```suggestion
    ).replace(/</g, '\\u003c')}</script>`;
```

justinfagnani: thanks!

PR #1386, opened 2022-11-08
Integrate wireit services and misc improvements

aomarks:
- Used (new!) Wireit services feature to streamline config of all scripts where relevant
- The top-level `npm run watch:dev` and `npm run watch:prod` scripts are now `npm start --watch` and `npm start:prod --watch`
- The `start` scripts now automatically restart servers whenever needed based on which files changed
- The `start` scripts now also launch the catalog server and firestore emulator

Also:
- Added some dependencies that were missing
- Fixed `tsc` errors by adding some new `build:types` scripts. 
- Fixed some format/lint errors
- Added CI jobs for `build` and `check`
- Switched the firestore emulator port to something more in line with the scheme we're using for the rest of the services
- Fixed a bug with `undefined` vs `null` that broke firestore calls
- Documented how to bootstrap firestore for local dev


packages/site-content/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "rootDir": "templates/src",
+    "noEmit": true,
+    "tsBuildInfoFile": "tsconfig.tsbuildinfo",
+    "module": "esnext",
+    "moduleResolution": "node",
+},
+  "include": ["templates/src/**/*.ts"],
```
justinfagnani, line 11: I guess disregard my previous comment, but... why isn't esbuild using the `include` setting here?

aomarks: I don't see a way to make esbuild pay attention to `include`. It always requires that you pass it the entrypoints explicitly, as far as I can tell.

package.json
```diff
@@ -12,8 +12,10 @@
     "build:dev": "wireit",
     "build:prod": "wireit",
     "start:docker": "docker-compose -f docker/docker-compose.yml up --build --remove-orphans",
-    "watch:dev": "npm run build:dev && (npm run start:dev > /dev/null 2>&1 & WIREIT_FAILURES=continue npm run build-and-check:dev --watch)",
-    "watch:prod": "npm run build:prod && (npm run start:prod & WIREIT_FAILURES=continue npm run build-and-check:prod --watch)",
+    "watch:dev": "npm run start-and-check:dev --watch",
```
justinfagnani, line 15: Do you think we need this pattern of a specific watch command or should we just use `--watch` on the command line?

packages/site-client/package.json
```diff
@@ -50,20 +51,31 @@
...
+      "files": [
+        "src/**/*.ts"
+      ],
+      "output": [
+        "lib/**/*.d.ts",
+        "tsconfig.tsbuildinfo"
+      ],
+      "dependencies": [
+        "../catalog-api:build",
+        "../custom-elements-manifest-tools:build"
```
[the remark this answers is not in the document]
aomarks: Yes

package.json
aomarks: That seems reasonable.

Actually how about this? `watch:dev` is now just `start`, and `watch:prod` is `start:prod`.

So to spin it up locally, you just write `npm start --watch`

.github/workflows/check.yml
```diff
@@ -0,0 +1,24 @@
...
+          cache: 'npm'
+          cache-dependency-path: package-lock.json
+
+      - uses: google/wireit@setup-github-actions-caching/v1
+
+      - name: NPM install
+        run: npm ci
+
+      - name: Check
+        run: npm run check
```
justinfagnani, line 24: why have separate workflows?

justinfagnani (approved):

.github/workflows/check.yml
aomarks: I like how they show up as different checks

PR #1389, opened 2022-11-22
Factor catalog server routes into separate files

justinfagnani:
This is in preparation for more catalog server routes.


packages/catalog-server/src/lib/server/routes/bootstrap-packages.ts
```diff
@@ -0,0 +1,84 @@
+/**
```
aomarks, line 1: Why `src/lib/server/routes` and not just `src/routes`?

justinfagnani: I dunno... my thinking was all the server parts in a folder with routes in a subfolder so that the files don't need a `-route` suffix.

aomarks (approved):

PR #1395, opened 2022-12-02
Deploy previews to cloud run when PR opened

aomarks:
Follow up to https://github.com/webcomponents/webcomponents.org/pull/1394

Pretty much the same as the `deploy-main.yaml` config, but doesn't update routing, and only allows up to 1 instance.

This is very similar to what we do on lit.dev, except this time I'm not including the SHA, so instead of getting a new preview URL for every commit, the same URL is updated on each commit instead.

Includes a GitHub Action which posts a comment to the PR with the URL.

Part of https://github.com/webcomponents/webcomponents.org/issues/1340
Part of https://github.com/webcomponents/webcomponents.org/issues/1339

### DIFF
--- a/.github/workflows/pr-preview-url-comment.yml
+++ b/.github/workflows/pr-preview-url-comment.yml
@@ -1,0 +1,35 @@
+# GitHub Action that creates a comment displaying the preview URL for each PR.
+name: PR Preview URL Comment
+
+on:
+  pull_request:
+    types: [opened]
+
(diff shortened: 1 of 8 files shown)


aomarks:
Hm, this didn't work. I'm going to try closing and re-opening the PR to see if that hits the cloud build trigger.

justinfagnani (approved):

PR #1398, opened 2022-12-09
Add command to export emulator data, import on start

justinfagnani:
This adds `emulators:export` to the catalog-server scripts. It can be run when the emulator is running to export data to catalog-server/firebase-data/. The `--import` flag on emulators:start works as long as that data folder exists, even if it's empty.

### DIFF
--- a/packages/catalog-server/package.json
+++ b/packages/catalog-server/package.json
@@ -19,6 +19,7 @@
     "test": "wireit",
     "test:manual": "wireit",
     "emulators:start": "wireit",
+    "emulators:export": "wireit",
     "firebase": "firebase"
   },
   "wireit": {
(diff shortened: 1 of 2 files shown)


packages/catalog-server/package.json
```diff
@@ -64,13 +65,21 @@
       "output": []
     },
     "emulators:start": {
-      "command": "firebase emulators:start --project wc-catalog",
+      "command": "firebase emulators:start --project wc-catalog --import firebase-data",
       "service": true,
       "files": [
```
aomarks, line 70: Was wondering the exported file should be in `files`, but I think not because the only effect of that would be to restart it when you export, which doesn't seem useful. (Not actionable, just noting for the record).

aomarks (approved):

packages/catalog-server/package.json
justinfagnani: That was my thought too. I didn't see how Wireit considering them input or output was useful.

PR #1400, opened 2022-12-10
Enable SSR for 11ty

### LINKED ISSUE #1337: [site-content] Integrate @lit-labs/eleventy-ssr

justinfagnani:
Fixes #1337

This doesn't integrate `@lit-labs/eleventy-plugin-lit` though, because we're sharing a template between 11ty and the server-generated catalog pages. We want to support content from 11ty which is a string, and content from server code which passes JS data into lit-html templates. So we want a lit-html template with a hole that can take unsafeHTML content or TemplateResults. The end result is much like `@lit-labs/eleventy-plugin-lit`


packages/site-server/package.json
```diff
@@ -53,6 +53,7 @@
       "command": "tsc --pretty",
       "files": [
         "tsconfig.json",
+        "tsconfig.tsbuildinfo",
```
aomarks, line 56: This should be an output, not an input

aomarks (approved):
The organization is sorta confusing to me, but hopefully we'll be able to chip away at some complexity over time -- maybe with some improvements to the SSR lib (e.g. the DOM shim timing issue?).

One blocking comment about the wireit config.

PR #1401, opened 2022-12-11
Add docs layout and page

justinfagnani:
This adds a WCOPage base class for page components and runs the 11ty generated pages through slots in page components. Seems to be a promising pattern for similar code organization between server and 11ty pages.


justinfagnani:
@aomarks this is ready for review now. It doesn't add docs, just a stand-in page to see the shell layout.

aomarks (approved):

PR #1404, opened 2023-01-06
[catalog-server] Handle invalid and not found packages better.

justinfagnani:
This handles a few cases better, and tests that the db state is what we expect:
1. Not found packages
2. Packages with no customElements field in package.json
3. Packages with a customElements field in package.json, but a missing file (something we see in the bootstrap packages)


packages/catalog-server/data/bootstrap-packages.json
```diff
@@ -1,40 +1,5 @@
 {
   "packages": [
-    "@shoelace-style/shoelace",
```
aomarks, line 3: Intentionally removed most of the bootstrap components?

packages/catalog-server/package.json
```diff
@@ -56,7 +56,7 @@
       ]
     },
     "test": {
-      "command": "NODE_OPTIONS=--enable-source-maps FIRESTORE_EMULATOR_HOST=localhost:7450 uvu test \"_test\\.js$\"",
+      "command": "NODE_OPTIONS=--enable-source-maps FIRESTORE_EMULATOR_HOST=localhost:7450 uvu test \"catalog_test\\.js$\"",
```
aomarks, line 59: Intentionally scoping down tests to just catalog ones?

packages/catalog-server/src/lib/catalog.ts
```diff
@@ -100,20 +103,31 @@ export class Catalog {
...
     }
-    console.log(' done');
 
     if (newPackage === undefined) {
-      await this.#repository.endPackageImportWithNotFound(packageName);
-      // TODO (justinfagnani): a crazy edge case would be a package that was
-      // previously found, but is not found now. Update package versions?
-      return {};
+      throw new Error(`Internal error importing package ${packageName}`);
+      // await this.#repository.endPackageImportWithNotFound(packageName);
```
aomarks, line 127: Remove commented out code?

aomarks (approved):

packages/catalog-server/package.json
justinfagnani: oops, I didn't mean to check that in

packages/catalog-server/data/bootstrap-packages.json
justinfagnani: nope, oops!

PR #1405, opened 2023-01-06
Compress manifest sources in the database

justinfagnani:
This lets us handle larger manifests.

I'm testing it with the largest manifest we've encountered so far: @lion/form-core.

I'm using Brotli from node:zlib assuming it's better at compressing JSON with strings that are likely to contain some markup than other methods. So far I'm using default options, but since this is a write-once, read-many operation, we could turn up the compression level.

Many of the changes here are not the actual compression itself, but getting the types setup to force us to compress and decompress at the right locations.


packages/catalog-server/src/test/lib/catalog_test.ts
```diff
@@ -358,4 +358,42 @@ test('Imports a non-existent package', async () => {
   assert.equal(importResult.packageInfo?.status, PackageStatus.NOT_FOUND);
 });
 
+test.only('Imports a large manifest', async () => {
```
aomarks, line 361:
```suggestion
test('Imports a large manifest', async () => {
```

packages/catalog-server/src/test/lib/catalog_test.ts
```diff
@@ -358,4 +358,42 @@ test('Imports a non-existent package', async () => {
...
+
+  const files = new LocalFsPackageFiles({
+    path: packagePath,
+    packageName,
+    publishedVersions: ['1.0.0'],
+    distTags: {
+      latest: '1.0.0',
+    },
+  });
+  const repository = new FirestoreRepository(TEST_SEQUENCE_THREE);
```
aomarks, line 375: How would I know what test sequence to use if I were writing a new test?

packages/catalog-server/src/lib/firestore/firestore-repository.ts
```diff
@@ -234,8 +248,13 @@ export class FirestoreRepository implements Repository {
       const distTags = packageMetadata['dist-tags'];
       const versionDistTags = getDistTagsForVersion(distTags, version);
 
+      const compressedManifest =
+        customElementsManifestSource &&
+        (await brotliCompress(customElementsManifestSource)).toString('base64');
+
       // Store package data and mark version as ready
       t.set(versionRef, {
+        __typename: 'ReadableCompressedPackageVersion',
```
aomarks, line 257: Is `__typename` something meaningful in firestore, or just a convention you picked? I see that it's a thing in graphql, but I would assume compression would be handled automatically by the firestore layer, so graphql shouldn't really be relevant?

aomarks (changes requested):

packages/catalog-server/src/test/lib/catalog_test.ts
justinfagnani: Good question, and "sequence" might be a bad name. The important part is that tests that rely on the same data need the same namespace so that one test can write data that another reads. That's why I named it "sequence" because all those tests need to run in order. But I'm using TEST_SEQUENCE_THREE incorreclty here because these don't belong to a sequence...

packages/catalog-server/src/lib/firestore/firestore-repository.ts
justinfagnani: __typename is a GraphQL thing and generated in the schema types by the graphql-codegen tools. It's optional, so we can use those GraphQL types as our internal data model too and generally ignore the __typename field. It'll be set by the GraphQL layers for the RPC calls.

By setting it to a literal that's different from 'ReadablePackageVersion' though, we force the ReadableCompressedPackageVersion to be unassignable to ReadablePackageVersion. I had tried to declare ReadableCompressedPackageVersion.customElementsManifest as `never`, but that didn't work.

aomarks (approved):

PR #1408, opened 2023-01-09
Add 11ty navigation plugin and some skeleton pages

justinfagnani:
Builds on https://github.com/webcomponents/webcomponents.org/pull/1401 to add a minimal docs outline.

There's a lot left to do: the outline isn't formatted well and includes section links that we want to remove. I'll add custom HTML output for the nav like we did for lit.dev.

### DIFF
--- a/packages/site-content/eleventy.config.prod.cjs
+++ b/packages/site-content/eleventy.config.prod.cjs
@@ -4,7 +4,10 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+const eleventyNavigationPlugin = require('@11ty/eleventy-navigation');
+
 module.exports = (eleventyConfig) => {
+  eleventyConfig.addPlugin(eleventyNavigationPlugin);
   eleventyConfig.addPassthroughCopy('site/assets');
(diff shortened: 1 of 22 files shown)


packages/site-content/eleventy.config.prod.cjs
```diff
@@ -4,13 +4,17 @@
...
 module.exports = (eleventyConfig) => {
+  eleventyConfig.addPlugin(eleventyNavigationPlugin);
   eleventyConfig.addPassthroughCopy('site/assets');
   eleventyConfig.addPassthroughCopy({'../site-client/bundled': 'js'});
   return {
     dir: {
       input: 'site',
       output: '_site',
     },
+    pathPrefix: '',
```
aomarks, line 18: Comment for this? I'm not sure why it's here (and why it's not in the `dev` config)

justinfagnani: oh, I can remove now. I was trying to get eleventyNavigation working from plain JS instead of Nunjucks and it was looking for the `url` filter. I fixed it by using the plugin API dynamically attached to the eleventyConfig object.

PR #1410, opened 2023-01-10
Render the element page description markdown as HTML

aomarks:


packages/site-server/package.json
```diff
@@ -103,12 +103,14 @@
     "@types/koa-conditional-get": "^2.0.0",
     "@types/koa-etag": "^3.0.0",
     "@types/koa-static": "^4.0.2",
+    "@types/marked": "^4.0.8",
```
justinfagnani, line 106: I remember a lot of considerations last time I picked a markdown library (things like supporting GH-flavor, using what 11ty does, the plugin API, etc). I'd like to run through and record these at some point, but whatever works now is great.

aomarks: I used the same one eleventy uses, that way we'll have consistency with our other markdown.

packages/site-server/src/lib/catalog/routes/element/element-route.ts
```diff
@@ -91,12 +100,32 @@ export const handleElementRoute = async (
     throw new Error('Internal error');
   }
 
-  const elementData = {
+  const declarationRef = parseReferenceString(customElement.declaration);
```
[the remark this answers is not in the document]
aomarks: Yeah was thinking that maybe 👍

PR #1412, opened 2023-01-12
Make wco-docs-page component and move nav HTML generation into it

aomarks:
- Make a wco-docs-page component
- Move nav HTML generation into wco-nav-page component
- Fix hydration of docs page

Also:
- Rename hydrate.js to hydrate-support.js
- Move hydrate-support directly into base template


packages/site-client/src/entrypoints/docs.ts
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright 2023 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import '../components/wco-docs-page.js';
+import type {NavEntry} from '../components/wco-docs-page.js';
+import {html} from 'lit';
+
+export const renderDocsPage = (navEntries: NavEntry[]) =>
+  html`<wco-docs-page .navEntries=${navEntries}></wco-docs-page>`;
```
justinfagnani, line 12: oops... we're missing the content here. This needs to be something like:

```suggestion
export const renderDocsPage = (content: unknown, navEntries: NavEntry[]) =>
  html`<wco-docs-page .navEntries=${navEntries}>${content}</wco-docs-page>`;
```

You can see the problem on the site at /docs/ where the content is missing.

justinfagnani (changes requested):

packages/site-client/src/entrypoints/docs.ts
aomarks: Fixed

justinfagnani (approved):

Issue #1416, opened 2023-01-13
Catalog search endpoint should be on site server


justinfagnani:
Right now the search feature hits an endpoint on the catalog server. This is fine if the catalog server is public or not proxied, but if the server it proxied it prevents the client from accessing the catalog server.

The difficulty is that the servers know their own URLs via env variables which point to the real server URLs (ie `https://pr1415---site-khswqo4xea-wl.a.run.app/`). But the local IAM proxy gives the servers URLs like `http://127.0.0.1:6453`. The client would have to know this localhost URL for the catalog server, which is unknowable from the server.

Having seach endpoints on the site server solves this and separates site public visibility from catalog public visibility.

PR #1417, opened 2023-01-13
Add search endpoint to site-server

justinfagnani:
This moves the search endpoint to the site server, removes the graphql client from the site client, and adds an integration test.

Fixes https://github.com/webcomponents/webcomponents.org/issues/1416

### DIFF
--- a/packages/site-server/src/lib/catalog/router.ts
+++ b/packages/site-server/src/lib/catalog/router.ts
@@ -6,10 +6,16 @@
 
 import Router from '@koa/router';
 import {handleCatalogRoute} from './routes/catalog/catalog-route.js';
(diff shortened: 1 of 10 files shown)


packages/site-server/src/lib/catalog/router.ts
```diff
@@ -6,10 +6,16 @@
 
 import Router from '@koa/router';
 import {handleCatalogRoute} from './routes/catalog/catalog-route.js';
+import {handleCatalogSearchRoute} from './routes/catalog/search-route.js';
 import {handleElementRoute} from './routes/element/element-route.js';
+// import cors from '@koa/cors';
```
aomarks, line 11: Remove commented lines

aomarks (approved):

PR #1419, opened 2023-01-15
Convert eleventy config and layouts to typescript

aomarks:


packages/site-content/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "rootDir": "./src",
+    "outDir": "./lib",
+    "tsBuildInfoFile": "tsconfig.tsbuildinfo",
+    "module": "CommonJS",
+    "skipLibCheck": false
```
justinfagnani, line 8: false is the default, no?

packages/site-content/package.json
```diff
@@ -18,31 +19,48 @@
...
+      ],
+      "dependencies": [
+        "../site-templates:build:types",
+        "../site-client:build:types"
+      ],
+      "clean": "if-file-deleted"
+    },
     "build:prod": {
-      "command": "eleventy --config=eleventy.config.prod.js",
+      "command": "eleventy --config=lib/eleventy.config.prod.js",
```
justinfagnani, line 40: I don't mind doing outDir='.' and gitignores to keep this file at the top level, fwiw.

packages/site-content/src/types/eleventy.d.ts
```diff
@@ -0,0 +1,39 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+declare module '@11ty/eleventy' {
+  export interface EleventyConfig {
```
justinfagnani, line 8: eleventy calls this class the `UserConfig`: https://unpkg.com/browse/@11ty/eleventy@1.0.2/src/UserConfig.js and they say you can get come typings with this:

```
/** @param {import("@11ty/eleventy").UserConfig} eleventyConfig */
module.exports = function (eleventyConfig) {
  // …
};
```
https://www.11ty.dev/docs/config/#type-definitions

Is it possible to just augment what types TS can infer on that import (if it supports a type of a JS module at all? I'm not sure), and should we align the name of the interface?

Also, here's what seems like a pretty nice declaration file if we want to cross check our definitions or import some, licenses permitting (it's MIT): https://github.com/chriskrycho/v5.chriskrycho.com/blob/main/types/eleventy.d.ts

packages/site-content/src/eleventy.config.dev.ts
```diff
@@ -4,11 +4,12 @@
...
-const fs = require('fs');
-const pathlib = require('path');
-const eleventyNavigationPlugin = require('@11ty/eleventy-navigation');
+import fs from 'fs';
+import pathlib from 'path';
+import eleventyNavigationPlugin from '@11ty/eleventy-navigation';
+import type {EleventyConfig, EleventyConfigResult} from '@11ty/eleventy';
 
-module.exports = (eleventyConfig) => {
+module.exports = (eleventyConfig: EleventyConfig): EleventyConfigResult => {
```
justinfagnani, line 12: Now that these are .ts files, can this be a default export?

justinfagnani (approved):

packages/site-content/tsconfig.json
aomarks: It was set to `true` in the base config. `catalog-server` doesn't compile with it enabled -- something to do with a library called `natural`. I've removed it from the base config and set it to `false` only in `catalog-server`.

packages/site-content/package.json
aomarks: Done

packages/site-content/src/eleventy.config.dev.ts
aomarks: If we write `export default` then the commonjs we get is:

```js
exports.default = (eleventyConfig) => ...
```

but that doesn't work with eleventy. we need:

```js
module.exports = (eleventyConfig) => ...
```

packages/site-content/src/types/eleventy.d.ts
aomarks: Adding the `@param` comment doesn't do anything at all for me. It just gets implicit `any`.

https://unpkg.com/browse/@11ty/eleventy@1.0.2/src/UserConfig.js  defines things like `addPlugin` on an interface called `Config`, and then there's a thing called `UserConfig` which strips off the methods.

So are you sure that's the thing that's passed as the `eleventyConfig` parameter, or is it supposed to be the thing you return with `dir` etc.?

I'm sort of confused and the name doesn't seem to matter very much, so I'll just leave things as they are for now.

justinfagnani: Huh... the location of that `@param` annotation says that it should be the type of the param, not the return. Given the view of types on the project, I wouldn't be surprised if that setup wasn't tested very well.

PR #1421, opened 2023-01-18
Add Import Package page to catalog

justinfagnani:
This adds a `/catalog/import` GET endpoint for the import page and POST endpoint for the import operation.

The page is extremely basic and just dumps the JSON output of the response to the screen.

Did a few cleanups on route file organization to make things simpler too.


packages/site-client/src/pages/import/wco-catalog-import-page.ts
```diff
@@ -0,0 +1,59 @@
...
+import {html} from 'lit';
+import {customElement, state} from 'lit/decorators.js';
+import {WCOPage} from '../../shared/wco-page.js';
+import {Task} from '@lit-labs/task';
+
+@customElement('wco-catalog-import-page')
+export class WCOCatalogImportPage extends WCOPage {
+  private _importTask = new Task(this, {
+    task: async ([packageName]: [packageName: string | undefined]) => {
+      console.log('_importTask', packageName);
```
aomarks, line 16: Remove log?

packages/site-client/src/pages/import/wco-catalog-import-page.ts
```diff
@@ -0,0 +1,59 @@
...
+        const response = await fetch(`/catalog/import`, {
+          method: 'POST',
+          headers: {
+            'Content-Type': 'application/json',
+          },
+          body: JSON.stringify({
+            packageName,
+          }),
+        });
+        const result = await response.json();
```
aomarks, line 28: Think we need to check the response code before parsing the json? Since fetch doesn't throw on HTTP errors.

packages/site-server/src/lib/catalog/routes/import-page.ts
```diff
@@ -0,0 +1,36 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
```
aomarks, line 3:
```suggestion
 * Copyright 2023 Google LLC
```

packages/site-server/src/lib/catalog/routes/import-page.ts
```diff
@@ -0,0 +1,36 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// This must be imported before lit
```
aomarks, line 7: Remove comment (seems like it's in the wrong place, and it's not true anymore anyway because of the new built-in SSR polyfilling)

packages/site-server/src/lib/catalog/routes/import-api.ts
```diff
@@ -0,0 +1,84 @@
...
+`;
+
+export const handleCatalogImportApiRoute = async (
+  context: ParameterizedContext<
+    DefaultState,
+    DefaultContext & Router.RouterParamContext<DefaultState, DefaultContext>,
+    unknown
+  >
+) => {
+  console.log('context.request.body', context.request.body);
```
aomarks, line 55: Remove log

aomarks (changes requested):
Needs a test? Lots of things could go wrong here, e.g. the graphql query is unvalidated I believe?

packages/site-client/src/pages/import/wco-catalog-import-page.ts
justinfagnani: I think we can let `.json()` throw and add a error template.

PR #1429, opened 2023-01-20
Trigger lit project board external issue webhook on new issue

aomarks:
Since we can't directly add external (different org) issues to a project board using the GitHub API, we instead do it with a playwright script that controls the UI.

This PR adds an action that will trigger the webhook whenever an issue is added to this repo.

See https://github.com/lit/lit/pull/3593 for more details.


aomarks:
@augustjk suggested we try using the `addProjectV2DraftIssue` API (https://github.com/lit/lit/pull/3593#issuecomment-1397792732), and it turned out doing that with the URL as the title totally worked. Will send another PR that does that instead.

PR #1430, opened 2023-01-20
Use addProjectV2DraftIssue API to add issues to lit board

aomarks:
Turns out you can set the `title` to an issue URL when using the `addProjectV2DraftIssue` API, and it does exactly what we want!

Need to merge this in order to test.


augustjk (approved):
Yay! This is great.

Unrelated to changes but wdyt about skipping the first API call of getting the project id and hard coding it into the workflow?

aomarks:
> Unrelated to changes but wdyt about skipping the first API call of getting the project id and hard coding it into the workflow?

It's sort of hard to discover the internal database ID, so I kinda thought it's nice to just look it up here so that the workflow is more maintainable.

PR #1434, opened 2023-01-20
Automatically add webcomponents.org area to project board issues

aomarks:
Also just hard-codes the project ID, and instead documents how you would look it up using the explorer.


augustjk (approved):
Nice!

PR #1436, opened 2023-01-20
Fix bugs with the query to add area to issues

aomarks:


.github/workflows/add-issues-to-project.yml
```diff
@@ -89,10 +90,10 @@ jobs:
           VALUE_ID: 9a4cb01d # webcomponents.org
         run: |
           gh api graphql -f query='
-            mutation ($project_id: ID!, $item_id: ID!, $field_id: ID!, $value_id: ID!) {
+            mutation ($project_id: ID!, $item_id: ID!, $field_id: ID!, $value_id: String!) {
```
augustjk, line 93: Huh, interesting. The docs do say this field should be `String`. Confusing field name.

augustjk (approved):

Issue #1444, opened 2023-03-01
[catalog] Add a package page


rictic:
The package page should show all of the elements in a package.

PR #1448, opened 2023-03-03
Add a basic package page.

rictic:
Fixes https://github.com/webcomponents/webcomponents.org/issues/1444

### DIFF
--- a/packages/site-client/src/pages/package/wco-package-page.ts
+++ b/packages/site-client/src/pages/package/wco-package-page.ts
@@ -24,12 +24,6 @@ export class WCOPackagePage extends WCOPage {
   static styles = [
     WCOPage.styles,
     css`
-      .full-screen-error {
-        display: flex;
-        flex: 1;
-        align-items: center;
-        justify-items: center;
-      }
       h1 {
         display: inline-block;
       }
@@ -49,7 +43,7 @@ export class WCOPackagePage extends WCOPage {
 
   renderContent() {
     if (this.packageData === undefined) {
-      return html`<div class="full-screen-error">No package to display</div>`;
+      return this.fullScreenError('No package to display');
     }
 
     return html`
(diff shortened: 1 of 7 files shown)


packages/site-client/src/pages/package/wco-package-page.ts
```diff
@@ -0,0 +1,74 @@
...
+  version: string;
+  elements: CustomElement[];
+}
+
+@customElement('wco-package-page')
+export class WCOPackagePage extends WCOPage {
+  static styles = [
+    WCOPage.styles,
+    css`
+      .full-screen-error {
```
justinfagnani, line 27: We should eventually move this to common styles

packages/site-server/src/lib/catalog/routes/package/package-route.ts
```diff
@@ -0,0 +1,95 @@
...
+    }`,
+  });
+
+  if (result.errors !== undefined && result.errors.length > 0) {
+    throw new Error(result.errors.map((e) => e.message).join('\n'));
+  }
+  const {data} = result;
+  const packageVersion = data.package?.version;
+  if (packageVersion === undefined) {
+    throw new Error(`No such package version: ${packageName}`);
```
justinfagnani, line 64: I think we should probably be rendering a 404 in these cases. TODO?

justinfagnani (approved):

packages/site-client/src/pages/package/wco-package-page.ts
rictic: +1, done

packages/site-server/src/lib/catalog/routes/package/package-route.ts
rictic: +1, TODO added

